PR #5, opened 2022-11-27
Atspi rustfmt.toml

luukvanderduim:
This PR contains two commits:

- The introduction of a `rustfmt.toml`
  The config introduces a few defaults and, most importantly perhaps, a link to the fustfmt options' documentation.
  The options that were chosen are stable.

- The crate's sources reformatted using the new config.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -107,7 +107,7 @@ dependencies = [
 
 [[package]]
 name = "atspi"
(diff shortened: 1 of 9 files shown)


luukvanderduim:
Feel free to disagree and play around with (stable) options and make suggestions, so we can land upon a unified layout.

PR #6, opened 2022-11-28
Remove async_trait; use Rust's native async traits instead

TTWNO:


TTWNO:
This pull request is awaiting the landing of async traits into stable. Once this happens, it should be fairly simple to merge.

PR #16, opened 2022-12-01
[draft] Adds `set_session_accessability`

luukvanderduim:
Set the `IsEnabled` property in the session bus.

 Assistive Technology provider applications (ATs) should set the accessibility
 `IsEnabled` status on the users session bus on startup as applications may monitor this property
 to  enable their accessibility support dynamically.

 See: The [freedesktop - AT-SPI2 wiki](https://www.freedesktop.org/wiki/Accessibility/AT-SPI2/)


TTWNO:
I get a warning on compilation.

```
warning: `must_use` attribute on `async` functions applies to the anonymous `Future` returned by the function, not the value within
   --> src/connection.rs:88:1
    |
88  |   #[must_use = "AT's are required to set `IsEnabled` on startup."]
    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
89  | / pub async fn set_session_accessibility(
90  | |     status: bool,
91  | | ) -> std::result::Result<(), zbus::Error> {
92  | |     // Get a connection to the session bus.
...   |
101 | |     Ok(())
102 | | }
    | |_- this attribute does nothing, the `Future`s returned by async functions are already `must_use`
    |
    = note: `#[warn(unused_attributes)]` on by default

warning: `atspi` (lib) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 34.30s
```

Also, what is `smol` for? I see it in Cargo.toml

luukvanderduim:
I've removed the `#must_use` attribute here because what it does is not what I intended.
That should fix the warning.

luukvanderduim:
`smol` provides the executor needed in the doctest.
The example in the documentation gets tested uses a `Future` and it needs an executor to complete.

try:
```Bash
cargo test --doc set_session_accessibility
```
Now, originally I tried using `futures::executor::block_on`, by adding futures as a requirement to `[dev-dependencies]`, where test dependencies reside, but as it appears you cannot specify the same crate in both `[dependencies]` and `[dev-dependencies]` differing in feature requirements. I could not get it to work. ( But I would love to be proven wrong! )

TTWNO:
Why did you close this? It seems like a good PR

luukvanderduim:
That was by mistake, sigh. What happened:
Synced my forks' main, pulled it into my local main, rebased my local feature branch to main and tried to push the changes but GitHub objected. My local branch had a different HEAD(?) So I synced the fork's feature branch, dropped the  the offending commit as I was about to push updates anyway this would not matter - so I thought.. (the mistake that purged the PR), pushed my local feature branch to GH, then found out the PR was gone. 

I opened a new PR: #17 
which is updated and has additional  comments.

PR #17, opened 2022-12-01
[draft] Add `set_session_accessibility` (again)

luukvanderduim:
Set the `IsEnabled` property in the session bus.

 Assistive Technology provider applications (ATs) should set the accessibility
 `IsEnabled` status on the users session bus on startup as applications may monitor this property
 to  enable their accessibility support dynamically.

 See: The [freedesktop - AT-SPI2 wiki](https://www.freedesktop.org/wiki/Accessibility/AT-SPI2/)

- rebased against main
- removed `#[must_use]`. It does something different than I intended. Useful, but not here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,3 +32,4 @@ zbus = { version = "^3.0.0", default-features = false }
 
 [dev-dependencies]
 byteorder = "1.4.3"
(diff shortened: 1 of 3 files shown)


TTWNO:
Makes sense.

PR #18, opened 2022-12-02
Trade `smol1` for `futures-lite` in dev-dependencies

luukvanderduim:
Reasoning:
`futures-lite` is that part of `smol` that provides the executor we need.

This reduces our footprint.

Before this commit:
```Bash
	cargo tree | wc -l
	239
```
After:
```Bash
	cargo tree | wc -l
	202
```


TTWNO:
Good work!

PR #19, opened 2022-12-02
Protocol implementation not bindings

luukvanderduim:
This PR :
  * makes `atspi` state it is a full protocol implementation instead of bindings.
  * fixes minor MD violation.


TTWNO:
Thank you so much for your contributions!

PR #20, opened 2022-12-07
Documentation and clippy fixes

luukvanderduim:
This addresses a number of problems Clippy noticed or silences Clippy where it was overzealous.
See the individual commits for further details


TTWNO:
Nice work adding docs and telling clippy we're fine!

PR #21, opened 2022-12-08
Add an include directive to minimize the published package size of `atspi`.

luukvanderduim:
Crates that added `atspi` as a dependency do not need all of the repository. Adding an [include or exclude directive](https://doc.rust-lang.org/cargo/reference/manifest.html?highlight=include%20directive#the-exclude-and-include-fields) allows for minimizing the projects published size.

This PR reduces the published version of atspi by 27% or 120.3 KB in 26 files (of 447.5 KB and 75 files in entire crate)

Achieved with the help of [cargo-diet](https://crates.io/crates/cargo-diet)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ homepage = "https://github.com/odilia-app/atspi"
 keywords = ["screen-reader", "accessibility", "a11y", "tts", "linux"]
 categories = ["accessibility", "api-bindings"]
 edition = "2021"
+include = ["src/**/*", "LICENSE-*", "README.md"]
 
 [package.metadata.release]
 release = true


TTWNO:
A lot of the files rely on eachother, so I'm not dead set on the idea.
I think this is a case where crate features might come in handy, so that deoendency files can be dealt wuth accordingly.

I have a draft branch up there if you'd like to start from there? Or, explain why this imclude/exclude directive is better than crate features.

luukvanderduim:
features and inclusion/exclusion directives serve separate goals and do different things.

Features:  Offer selective inclusion of pieces of code to the downstream user of your crate.  "I only want feature X and Y but do not require Z"
Features reduce compile time because the compiler does has less work to. Does not need to compile everything and does not need to optimize away unused code at link time.

'Include / exclude directives´: Allow the crate maintainer to select what parts of the crate are relevant to be published on `crates.io`.  What is published is mostly used by compilers and for humans when we use 'go to definition' in our IDE's.

Latest `atspi` ended up here, the hash after github.com-* is most likely different in your registry. 
```Bash
~/.cargo/registry/src/github.com-1ecc6299db9ec823/atspi-0.3.9
```
Humans rarely visit this directory. We do not need nor want the `xml/` files or `.gitignore` here for instance. Those are well kept on our GitHub repository where we keep them for future reference.

Generally, when people want your code, they will clone your repo and look there for extras.
This is also where people go if they want to collaborate in some way.

But event if `odillia-app` ceases to exist, there will always be copies of each published version on `crates.io` so people can compile against each published version of ` atspi` if they desire.

You can see what will be included in the published crate by: 
```Rust
cargo package --list
```
I think it is good practice to specify specifically what goes in the published registry as you have limited control over it after publishing and you can yank published crates because you accidentally published something, it is better to prevent it.

I think it is also a good idea to minimize what is published to the crates.io registry to minimize what users need to pull.

Just google around for what you should / should not publish on crates.io and you might want to read a bit of:

The section on ['the include and exclude field](https://doc.rust-lang.org/cargo/reference/manifest.html?highlight=include%20exclude%20directive#the-exclude-and-include-fields) from the Cargo book.

TTWNO:
Fantastuc explaination! Sorry, I guess I didn't have my coffee this mornimg.

luukvanderduim:
Thanks! 
Do ask if there is something unclear. I learn as we go too. 
Knowing someone will read the commit messages already makes me think more carefully about the commits' justification.

PR #24, opened 2022-12-12
Change `Event` validation on messages to `MessageType::Signal` test.

luukvanderduim:
Signals are broadcasted Events. So, to test for events, it is sufficient to test if the `Message` is `MessageType::Signal`. This test is likely a little bit cheaper than the string compare, but that is not the point. Point is that we allow 5 more streams to come through `RegistryProxy`:

$ rg ' signal' src/events/ -trust | wc --lines
56

$ rg ' signal' src/ -trust | wc --lines
61

The difference is made up by these events:

src/registry.rs
31:    /// EventListenerDeregistered signal
35:    /// EventListenerRegistered signal

src/cache.rs
59:    /// AddAccessible signal
63:    /// RemoveAccessible signal

src/socket.rs
30:    /// Available signal

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1,7 +1,7 @@
 use std::ops::Deref;
 
 use futures::stream::{Stream, StreamExt};
-use zbus::{Address, MessageStream};
+use zbus::{Address, MessageStream, MessageType};
 
 use crate::{bus::BusProxy, events::Event, registry::RegistryProxy};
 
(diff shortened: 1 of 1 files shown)


TTWNO:
This is good stuff! Can you confirm that other signal types are able to be received or is this hypothetical?

luukvanderduim:
I think we should drop this PR - it has served its purpose! ;)

PR #27, opened 2022-12-15
Fully typed events

TTWNO:
Starting with a Python script to auto-generate our Rust enums/structs for events.

For those wondering, `zbus-xmlgen` doesn't do exactly what we want. It creates a separate enum for each interface with the `#[dbus_proxy]` macro. This is to lay out strongly-typed events; let's only serialize data which has a specific name in each XML declaration.

Next step is to serialize to these various events using pattern matching, and maybe some constants. "org.a11y.atspi.Event.Object" should be some kind of constant associated with either ObjectProxy, or some other constant in a file.


DataTriny:
@TTWNO  I like the approach you're taking with the Python script. However, don't forget that replacing every occurrence of a type by another one won't work everywhere. Take `Accessible::get_state` for instance.

PR #28, opened 2022-12-15
[Luuk's] Fully-Typed Events

luukvanderduim:
- `#[must_use]` attribute indicates that the user or the method must do something with the return type.
- Functions that Result<> should document when the functon `Err`'s

You could also choose to tell Clippy not to complain about it, but either way, I find the squigllies distracting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 /target
 Cargo.lock
+/atspi-macros/target
+/atspi-matcros/Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ homepage = "https://github.com/odilia-app/atspi"
 keywords = ["screen-reader", "accessibility", "a11y", "tts", "linux"]
 categories = ["accessibility", "api-bindings"]
 edition = "2021"
-include = ["src/**/*", "LICENSE-*", "README.md"]
+include = ["src/**/*", "atspi-macros/**/*", "LICENSE-*", "README.md"]
 
 [package.metadata.release]
 release = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,11 +23,14 @@ gvariant = ["zbus/gvariant"]
 tokio = ["zbus/tokio"]
 
 [dependencies]
+atspi-macros = { path="atspi-macros"}
 async-recursion = "^1.0.0"
 async-trait = "^0.1.59"
 enumflags2 = "^0.7.5"
 futures = { version = "^0.3.25", default-features = false }
+quote = "1.0.21"
 serde = { version = "^1.0", default-features = false, features = ["derive"] }
+syn = "1.0.105"
 tracing = "^0.1.37"
 zbus = { version = "^3.0.0", default-features = false }
 
--- a/atspi-macros/Cargo.toml
+++ b/atspi-macros/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "atspi-macros"
+version = "0.1.0"
+edition = "2021"
+authors = ["Luuk van der Duim <luukvanderduim@gmail.com>"]
+description = "Macros to assist in creation of the lunanode crate."
+license = "MIT"
+readme = "README.md"
+keywords = ["macros", "helper"]
+
+[lib]
+proc_macro = true
+
+[dependencies]
+proc-macro2 = "1.0"
+syn = { version = "1.0", features = ["full"] }
+quote = "1.0"
+zbus_names = "2.4.0"
--- a/atspi-macros/src/lib.rs
+++ b/atspi-macros/src/lib.rs
@@ -43,7 +43,11 @@ pub fn try_signify(input: TokenStream) -> TokenStream {
             }
         }
 
-        impl Signified for #name {}
+        impl Signified for #name {
+            fn properties(&self) -> &HashMap<String, OwnedValue> {
+                self.inner().properties()
+            }
+        }
     };
 
     // Return the expanded code as a token stream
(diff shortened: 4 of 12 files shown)


luukvanderduim:
Adds AtspiEvent enum and relevant TryFrom and select-upon-signature-logic.

I have not yet tried what the script in `xml/` will do, hopefully it will produce the many event-innate types for us.

TTWNO:
The python script should generate structs for every event type. Feel free to chsnge the format inside the script if you like one way or another better

TTWNO:
Most of it looks really good, but the dedicated streams for varuous types for sure has to go. As discussed in the issue thread, it maks sense for clinets to use a second connecting if they want separate ateeams.

Gonna need more time to integrate with Odilia and see how it works in practice, but my initial thoughts are very positive.

luukvanderduim:
Thanks.
I would like to make sure we are on the same page though before I start deleting things.

In the previous discussions we settled on merging both ideas: 
Having types that express in type the meaning of the event. So no more looking up what `any_data` means in the context of that particular event. Excellent idea!

I think it is sensible to have the  
- The streams that offer the un-opinionated, homogeneous, pure bus types.
(fast and flexible)

as well as - later on- offer more opinionated and convenient:

- Streams that build upon those:
-  Qt/Atspi (with Qspi signals translated to Atspi)
-  All events stream
-  Cache events stream (heterogenous, may return an enum(

There may be even Qt developers who would like a stream wherein atspi events are converted to Qspi types and stream that. No idea why, but library authors' own custos should not hamper what others might want.

Those in trait(s on Connection, maybe feature gate them or offer those in a sub-crate.. I am thinking aloud at this point.

But admittedly I need to finish the homogeneous  ones first. Unless you oppose those?

luukvanderduim:
I think it is also just a matter of trying and see what works well and what less so.

TTWNO:
Our higher level types are already wrappers around the base Event types. This already gives us all the info we could possibly need. Multiple streams doesn't make sense, since the user program could just make a second connection or filter it themselves and divvy it out to various channels, queues, etc.

luukvanderduim:
Ill make it one stream.

TTWNO:
Finally had an opportunity to look at this more closely. I think the vast majority of it looks great! It for sure solves the problem it's meant to solve. Odilia seems to throw around 15 errors after changing to this new library, which is pretty small in the larger picture.

Your try_from implementation for each event type could use some work. The happy-path of the code should not be burried within an if/else block if possible. Take for example the function `AtspiEvent::try_from`:

```rust
    fn try_from(message: Arc<Message>) -> Result<Self, Self::Error> {
        let body: EventBodyOwned = match message.body_signature() {
            Ok(sig) => {
                if sig == connection::QSPI_EVENT {
                    EventBodyOwned::from(message.body::<EventBodyQT>()?)
                } else {
                    message.body::<EventBodyOwned>()?
                }
            }
            Err(e) => return Err(AtspiError::from(e)),
        };
        Ok(Self { message, body })
    }
```

This could be written like this, which is quite a bit more readable for the next maintainer of this section of code:

```rust
        let signature = match message.body_signature() {
          Ok(sig) => sig,
          Err(e) => return Err(AtspiError::from(e)),
        };
        let body = if signature == connection::QSPI_EVENT {
          EventBodyOwned::from(message.body::<EventBodyQT>()?)
        } else {
          message.body::<EventBodyOwned>()?
        };
        Ok(Self { message, body })
```

It's easier to keep track of a function's state when there is less state to think about. It'd be even better if the `Signature` type supported match arms, but since it doesn't, this would be a bit more concise way of doing it. Always define the error as high up as you can. I think the other `*Event::try_into` functions could use similar treatments.

The code looks really good though. Some missing types here and there, but it's really starting to take shape!

TTWNO:
One concern I have is that rhe signatures are not brought from the XML. Is it possible to write that as a macro?

luukvanderduim:
> Finally had an opportunity to look at this more closely. I think the vast majority of it looks great! It for sure solves the problem it's meant to solve. Odilia seems to throw around 15 errors after changing to this new library, which is pretty small in the larger picture.

Yes, the event it is looking for has a new type `AtspiEvent` instead of `Event` and you have to handle the event enum now.

> 
> Your try_from implementation for each event type could use some work. 
The happy-path of the code should not be burried within an if/else block if possible. Take for example the function `AtspiEvent::try_from`:
> 
> ```rust
>     fn try_from(message: Arc<Message>) -> Result<Self, Self::Error> {
>         let body: EventBodyOwned = match message.body_signature() {
>             Ok(sig) => { 
>                 if sig == connection::QSPI_EVENT {
>                     EventBodyOwned::from(message.body::<EventBodyQT>()?)
>                 } else {
>                     message.body::<EventBodyOwned>()?
>                 }
>             }
>             Err(e) => return Err(AtspiError::from(e)),
>         };
>         Ok(Self { message, body })
>     }
> ```

I did not touch this one (much). I renamed the `Event` -> `AtspiEvent` and adjusted the associated `Error` type.
I will gladly apply your suggestions, because I fully  agree.


> Since you've implemented `From<zbus::Error>` on `AtspiError`, you can just use the question mark.
> 
> It's easier to keep track of a function's state when there is less state to think about. It'd be even better if the `Signature` type supported match arms, but since it doesn't, this would be a bit more concise way of doing it. Always define the error as high up as you can. I think the other `*Event::try_into` functions could use similar treatments.

No problem. Will treat those accordingly.

> 
> The code looks really good though. Some missing types here and there, but it's really starting to take shape!

Thank you!

luukvanderduim:
> One concern I have is that the signatures are not brought from the XML. Is it possible to write that as a macro?

Absolutely we could!  Would we perhaps like to extract a lot more from the XML at compile time aside from the signatures? Think of the string type identifiers for eg.: interfaces, paths.. 
But then where do we end. I know this project started out with dynamically generated interfaces (atspi-codegen).
I do request to make that a separate PR, after this one.

TTWNO:
Totally agee thst self generation should be a separate PR

TTWNO:
Merged this into a branch on the repo with some of my changes. You should have write privs now. "With great power..." and all that.
I'll link it to the new PR in a sec here. Let's work on this together.

PR #30, opened 2022-12-24
Fully typed events

TTWNO:
Following in the tradition of Luuk's fully-typed-events https://github.com/odilia-app/atspi/pull/28

Now adding Luuk as a contributor so we can work together on improving this branch.


luukvanderduim:
Before un-drafting:
- [ ]  `TryFrom<Into<AtspiEvent>>` for the enum types in `identify`

Associated tasks that could be done after merge:
- [ ] Semantically "Fine-type" user facing types.
- [ ] Add tables and docs to enums: `WindowEvents` and `ObjectEvents`
- [ ] Give each specific event a small description.
- [ ] Correct and add usage examples for docs.
- [ ] Exclude doctests on examples that can't (because they need a bunch of boilerplate that is not there.)

luukvanderduim:
I will implement the first of those 2022-12-26, if you can think of more things to do before merge, do not hesitate to add, suggest or reorder.

atspi-macros/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    // Generate the expanded code
+    let expanded = quote! {
+                #item_struct
+        impl TryFrom<Arc<Message>> for #name {
+            type Error = AtspiError;
+
+            fn try_from(message: Arc<Message>) -> Result<Self, Self::Error> {
+                                if message.member() != Some(MemberName::from_static_str(#member)?) {
+                                    return Err(AtspiError::CacheVariantMismatch);
```
luukvanderduim, line 128: We likely want a more generic `Error` variant here.
Along these lines perhaps?: `Error::MemberMatch(#name.to_string())`

TTWNO: Yeah this would be a MemberMismatchError, I suppose?

xml/meson.build
```diff
@@ -0,0 +1,32 @@
...
+  'Text.xml',
+  'Value.xml',
+]
+
+introspection_generated = custom_target(
+  'introspection_generated',
+  input: introspection_sources,
+  output: [ 'introspection.c', 'introspection.h' ],
+  command: [ generator, '@INPUT@', '--c-output=@OUTPUT0@', '--h-output=@OUTPUT1@' ],
+)
```
luukvanderduim, line 32: What does this do?

TTWNO: Lol! Oops. Looks like I copied it in from at-spi2-core by accident.

xml/generate-introspection.py
```diff
@@ -0,0 +1,100 @@
...
+}
+ARG_IDENT_FROM_NUMBER = [
+  "kind",
+  "detail1",
+  "detail2",
+  "any_data",
+  "properties"
+]
+
+# will store lists of members with their various details. Will output a markdown table.
```
luukvanderduim, line 21: Thank you! .

atspi-macros/src/lib.rs
luukvanderduim: Sounds good to me!

TTWNO:
Looks really good!

The `Event::Atspi` variant should contain an additional enum layer: `AtspiEvents`, and this should contain all sub-categories by interface: `Document`, `Accessible`, `Window`, etc. The innermost type will implement `GenericEvent`.

This works best for pattern matching.

We'll make all this code auto-generate on the next PR, let's just get something working for now.

luukvanderduim:
Sorry, but we are not ready just yet. Please try the test app:

github.com:luukvanderduim/a11y-bus-cache-events.git

There is also still a `CacheVariant` error that should be a MemberMatch thing in the proc-macros. I tried to fix it but my efforts were fruitless. Debugging from proc-macro context is a bit of a pain, I would say.

TTWNO:
I've merged this for now, since it seems to work and not error. I added some additional items to the macros.

Since we're not 1.0 yet, I'm fine breaking things for a while. Feel free to open new issues, or to commit yourself since you have access now (I'm the only one who can publish to crates.io for now though).

Great work!

PR #42, opened 2023-01-05
State <-> &str

TTWNO:


DataTriny:
Before I review this, @TTWNO can you please describe the use case you have in mind? Can it be solved by deserialization instead?

TTWNO:
It *should* be solved be (de)serialization, but as far as I understand it, the data beeds to be in some sort of format for serde to serialize it. I may be wrong, but I don't beleive you can (de)serialize just a plain enum value with serde without using some kind of trick to turn it into a JSON/XML/TOML string.

If you know of a way to do this with serde, I'd *much* prefer that.

DataTriny:
I already added everything in a previous PR:

```rust
...
#[derive(...Serialize, Deserialize...)]
#[serde(rename_all = "kebab-case")]
pub enum State {
```

And that's what I use to send `org.a11y.atspi.Event.Object` `StateChanged` signals in AccessKit.

TTWNO:
Are you saying we should bring in serde_plain to handle the case of mapping a &str to a State and vice-versa?

TTWNO:
I tried that out, and it seems to work. You're right, this is a better solution. Closing this now.

PR #44, opened 2023-01-05
Fix CI issues

TTWNO:
Some concerns:

1. I'm getting "module name reiterated as suffix" sort of clippy error on all the Enums in identify.rs; this is because the module name is separate to distinguish between object::PropertyChangeEvent and window::PropertyChangeEvent. Should I ignore these errors, or is there some better way to format that file (note the `identify.rs` is auto-generated, so please suggest the change here and I will put it in `atspi-codegen`).
2. Some methods have "too many parameters", but those parameters have been inserted by the [dbus_proxy](https://docs.rs/zbus/latest/zbus/attr.dbus_proxy.html) macro, not the code itself; it would never be visible to end-users. Should this be ignored as well?

That's just about it, I'm pretty happy with how close we are to passing clippy here.


luukvanderduim:
Wirh respect to concern no.1:
[module-name-repetitions](https://rust-lang.github.io/rust-clippy/master/#module_name_repetitions) is a 'pedantic' lint to prevent users to type the same twice. As such it is fairly harmless.
We use 'Events' as suffix on the enums to indicate this groups events.

Three possible fixes: 
1. change module name
2. #![allow(clippy:module_name_repetitions)]
3. change the names and refactor all references accordingly

Personally, I lean towards no. 2 as it acknowledges the issue and handles it by saying it is fine - but option 1. is also fine.
Option 3. may be more involved.

With respect to 2. 'too many parameters': 
We have no say in what DBus interfaces look like. If a member has ten arguments, it has ten arguments. 
Also here, I lean towards acknowledge the issue, comment on it, and silence clippy on the issue.

TTWNO:
@luukvanderduim , I've added tests to the CI, so it might be good if you could add a test for the event_stream function. How to do that, I'm not totally sure. I don't think anything will be picked up in a container like that.
Is there a way to mock accessibility events for the sake of the test?

luukvanderduim:
I hope the container has a bus. If at-spi2-core I think `launcherd` to be there too. Then we can launch the a11y bus, if it did not already.

Then we get the accessibility bus:
```bash
busctl --user call org.a11y.Bus /org/a11y/bus org.a11y.Bus GetAddress
```
We may get a long way using [FreeDesktop's `busctl` capability to emit signals](https://www.freedesktop.org/software/systemd/man/busctl.html#emit%20OBJECT%20INTERFACE%20SIGNAL%20SIGNATURE%20ARGUMENT).
```bash
busctl --address unix:path=/run/user/1000/at-spi/bus_0 emit 
```

have a [CI test(s)](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-directory) in atspi/tests/ to simulate our crate users.

In such a test we could receive signals we send with busctl

We can hopefully send signals with commands along these lines:

`busctl --address unix:path=/run/user/1000/at-spi/bus_0 emit /org/a11y/atspi/Event/Mouse org.a11y.atspi.Event.Mouse siiv\(so\) "Button" "p1" --destination=org.freedesktop.DBus.Registry `

For more complex testing I will need to write a test application to receive method calls and emit predefined signals back.
Not tested it but this is what I come up with now.

TTWNO:
I have a test that can receive events, but it's on my desktop. I'll push it this afternoon.

TTWNO:
The test runs forever, but technically it's in there now. Now to figure out how to send a proper event via `busctl emit`

TTWNO:
We should probably auto-generate some of these tests. For example, it might be nice to test for all possible event types based on what's in the XML, then check that those values got passed in correctly. But I think this PR is good enough for now. I'll make a separate one some time with those more advanced features.

src/connection.rs
```diff
@@ -74,7 +74,63 @@ impl Connection {
     ///
     /// # Example
     /// ```
-    /// todo!()
+    /// use zbus::{fdo::DBusProxy, MessageType, MatchRule, Address};
+    /// use futures_lite::future::{block_on, race, yield_now};
+    /// use futures_lite::pin;
+    /// use futures_lite::StreamExt;
+    /// let receive_good_event = async {
+    ///    let connection = atspi::Connection::open().await.expect("Could not open a11y bus.");
```
luukvanderduim, line 82: You can [hide parts of the example](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example) (eg. boiler plate) from from the reader by using '#' .

src/connection.rs
```diff
@@ -74,7 +74,63 @@ impl Connection {
...
+    ///    let object_match_rule = MatchRule::builder()
+    ///        .msg_type(MessageType::Signal)
+    ///        .interface("org.a11y.atspi.Event.Object").expect("Can not build MatchRule with interface org.a11y.atspi.Event.Object")
+    ///        .build();
```
luukvanderduim, line 86: Maybe just `unwrap()` in examples. I have recently [come to learn (std lib docs)](https://doc.rust-lang.org/stable/std/result/enum.Result.html#recommended-message-style) that the  `expect()` message should say why we expect the Result<> should be Ok. (why is it justified to call expect in this situation)
There generally is too little context or relevance for that in an example, plus, a simple unwrap is shorter.

src/connection.rs
```diff
@@ -74,7 +74,63 @@ impl Connection {
     ///
     /// # Example
     /// ```
-    /// todo!()
+    /// use zbus::{fdo::DBusProxy, MessageType, MatchRule, Address};
```
luukvanderduim, line 77: This test should be a [CI test](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-directory). 
What you are doing is trying to fit a CI test in a doc-test.
From the rust crate perspective, there are doc-tests (check if the doc example sttill does what it says it does), unit tests `#[test]` annotated, these test functions, mostly I believe. These live in the source code itself or are a module or if there are many, a mudule directory `atspi/src/tests/`, and lastly are integration tests (CI). All test- types are tested using cargo test.
 
The CI test is external to the crate (see the link above for more) but associated nevertheless as it lives in `atspi/tests/`. Cargp will compile each file there as separate program and run it as test.

This avoids lots of code here in connection and, more importantly, really links `atspi` as our users would.

src/connection.rs
```diff
@@ -74,7 +74,63 @@ impl Connection {
...
+    ///    pin!(a11y_event_stream);
+		///#   let addr_parts = std::process::Command::new("busctl").arg("--user").arg("call").arg("org.a11y.Bus").arg("/org/a11y/bus").arg("org.a11y.Bus").arg("GetAddress").output().expect("Could not get a11y bus address");
+		///#   assert_eq!(addr_parts.status.code().expect("No status code for `busctl` command"), 0, "Status code for first `busctl` command is not 0.");
+		///#   let addr_parts_string = String::from_utf8(addr_parts.stdout).expect("Output of `busctl` command is not in UTF-8 format");
+		///#   let mut addr_iter = addr_parts_string.split_whitespace();
+		///#   addr_iter.next(); // this is to remove the first "s" in the output.
+		///#   let mut addr_str = addr_iter.next().expect("Unable to get a11y bus address from output of `busctl` command").replace("\"", "");
+		///#   let mut create_event_base_cmd = std::process::Command::new("busctl");
+		///#   let mut create_event_cmd = create_event_base_cmd
+		///#       .arg("--address")
```
luukvanderduim, line 100: Could we call `busctl` as part of the Actions script?

luukvanderduim (changes requested):
You can move the test over to `atspi/tests` as further outlined in the review and link.

src/connection.rs
TTWNO: What is an "Actions script"?

src/connection.rs
TTWNO: I disagree, this is exactly how you should use the API, and it's a good example. We just need a way to not display the vast majority of the function due to the overhead of running a command through the shell.

We should absolutely have a doc test for something like this.

But we should *also* automate some CI tests which test different kinds of tests.

src/connection.rs
TTWNO: I can change it to unwrap.

src/connection.rs
luukvanderduim: I am not too well versed in the GitHub `Actions` lingo, but I meant the script we use to have our CI tests run.

src/connection.rs
luukvanderduim: > I disagree, this is exactly how you should use the API,

Not sure I follow you. What exactly do you disagree on?

> and it's a good example. 

I am not opposing the example, quite the opposite. And if a simple example can double as unit test, that is perfect.

> We just need a way to not display the vast majority of the function due to the overhead of running a command through the shell.

Well you are using the hash `#` there, it should work, but may need a preceding space. 
You can check what it brews by `cargo doc --open`.

> 
> We should absolutely have a doc test for something like this.
> 
I want an example, of course - and it should be doc-tested, but I am not sure the test you are doing here is a good fit. if you want integragion tests, tests that mimics our users and actually links to our library and receive something over a bus, then rusts [integration testing](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests) is perfect.

Imagine you want 57 tests, for every ATSPI end type one and some for Cache and some for the LitenerEvents, you do not want them here.

> But we should _also_ automate some CI tests which test different kinds of tests.

TTWNO: I'm saying that this is a good doc test, it just has a lot of boilerplate. Showing how to use the event_steam() return type is perfect here, and it's what I would expect from other libraries. Even if the test itself is quite long due to the requirement of sending out a signal afterwards.

TTWNO:
You removed the method to send an event on the bus from the example inside `src/connection.rs` which leads to a "hang". It's not really hanging, just waiting infinitely.

Fix incomming.

luukvanderduim (approved):

TTWNO:
Based on the PR, this does not introduce any incompatibilities so we'll merge this now.

PR #46, opened 2023-01-05
Add additional methods to BitFlags<State>

TTWNO:
I should probably add all those other methods too from [enumflags2](https://docs.rs/enumflags2/latest/enumflags2/struct.BitFlags.html) so that it is truly transparent.

Please await further commits


TTWNO:
> Ideally such a commit would add a test too.

I'm just passing through info to BitFlags, not really doing any original work. Unless `enumflags2` is having issues, we shouldn't be either.

src/state.rs
```diff
@@ -271,6 +276,31 @@ impl StateSet {
     pub fn iter(self) -> impl Iterator<Item = State> {
         self.0.iter()
     }
+
+		/// Create a StateSet, with all states active.
+		pub fn all() -> Self {
+			StateSet(State::all())
+		}
+
+		/// Checks if all states are set.
```
DataTriny, line 285: I don't see any use case for this. The purpose of creating newtypes is to guide people into using the API correctly, as well as giving them the tools that make sense. A `StateSet` is not a `BitFlags`. Therefore not all methods of `BitFlags` are applicable to `StateSet`.

src/state.rs
```diff
@@ -271,6 +276,31 @@ impl StateSet {
     pub fn iter(self) -> impl Iterator<Item = State> {
         self.0.iter()
     }
+
+		/// Create a StateSet, with all states active.
```
DataTriny, line 280: I am not sure why would anyone need this.

DataTriny (changes requested):

src/state.rs
TTWNO: Even just as a debug mechanism, showing a number or a bitset would be a lot less clear than a list of the states.

TTWNO:
I've removed some of the bitset methods that don't make as much sense, like is_all, and all(), but I've left the rest since the usage of those seems fairly clear.

src/state.rs
```diff
@@ -344,6 +374,41 @@ impl std::ops::BitXor for StateSet {
         StateSet(self.0 ^ other.0)
     }
 }
+impl std::ops::BitXorAssign for StateSet {
```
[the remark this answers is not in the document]
DataTriny: Can you please add a blank line between each `impl` block?

src/state.rs
```diff
@@ -262,6 +262,11 @@ impl StateSet {
         self.0.contains(other)
     }
 
+		/// Removes a [`State`] (optionally) previously contained in the [`StateSet`].
```
DataTriny, line 265: I don't like having links pointing to the current item inside its documentation. You already know you are reading the documentation for `StateSet`. Therefore it's only noise, especially for screen reader users.

DataTriny (changes requested):

DataTriny (approved):

PR #48, opened 2023-01-10
atspi-codegen reinstate-docs

luukvanderduim:
### atspi-codegen reinstate-docs
    
This adds functionality to save doc-comments before regenerating new sources and 
reinstate docs to `identify` afterwards. 
```shell
$ identify --help

Usage: identify [-s] [-o] [-r] [-f] [-i]

Save manual doc-comments, then generating new sources and reinstate manual doc-comments.

Options:
  -s, --docs-file   save manual doc-comments, then exit
  -o, --docs-stdout write manual doc-comments to stdout, then exit
  -r, --regen-stdout
                    regenerate sources from xml, write to stdout
  -f, --regen-file  regenerate sources from xml, write to source file
  -i, --insert      reinstate - restore docs from file
  --help            display usage information
```

It still has rough edges. It may well eat your pet.
That being said, it is the first version with this feature to be somewhat useful.

TODO:

* [x] Move mock-up event to a test tree.
* [x] Add tests for the 'reinstating comments' logic. Demonstrate reinstating is somewhat useful, otherwise sunset the whole idea.
* [x] Reinstate tables and other useful comments.
* [x] Merge with main again.


TTWNO:
Wow! That code generation just got a lot more complicated. I'll have to walk through how this works at this point.

What would be the steps to regenerate with the documentation that is already in the file. I'm not sure I understand the options. Would I do `identify -r ./src/identify.rs`, then `identify -i`?

I love the idea though, let's see if we can make this work.

luukvanderduim:
Glad you like it!
As of present the command line options are simply switches. No further values, such a path are expected.
The binary is intended for `identify.rs`, as it was. 
That being said, I kept in mind that you would want to use this as a template for other regeneration utilities so its file name can be found at the top of main. It is easy to make it accept a path. We can go either way, this is just to get something going.

You would use either `-r` or `-f` they are essentially the same but '-r' writes to stdout, '-f' writes to the source file.
(I think of the stdout options as 'dry-run')

Generally I imagine one would:

```shell
# to save the docs / comments:
identify -s 

# to save newly generated source:
identify -f

# (try to) reinstate docs and comments:
identify -i  
```
After each you have the opportunity to see if the result is what you expect.

> Wow! That code generation just got a lot more complicated. I'll have to walk through how this works at this point.

Code generation is left unchanged. It utilizes the functions you wrote.
The comments should help you, future me and perhaps others around.

TTWNO:
I understand the actual code generation is the same, I just mean there's more moving parts now.

TTWNO:
After attempting to add a comment to the ObjectEvents enum, then running your set of commands: `identify -r`, then with `-s` then with `-i` does not work. It moves all comment ignore blocks to the top and does not preserve comment content.

luukvanderduim:
Hmm. Tested it with the last known identify with all the docs. It saved and reinstated a bunch of docs, but it seems it is still not working well, sorry.
Not ok. I am on it.

luukvanderduim:
Made lots of fixes, mostly just false assumptions and an unexpected interaction with rustfmt.
Please try again!

luukvanderduim:
Comments are reinstated in the wrong place. Fixing that.

luukvanderduim:
This simplifies where docline(s) should be reinserted:

Previously it would keep track of what it had inserted and add that to the index.
That proved to be a shaky strategy.
Realizing it could just look up the correct index in the created `Vec<String>` 'source_and_doc_lines' which seems to work much better.

The end result is still a bit too spacious, but a `cargo fmt` solves that too.

To demonstrate what it does, I reinstated docs in a newly generated `identify.rs`.
Like so:
```rust
identify -f # created a new 'identify'
identify -i # reinstates docs from commit: https://github.com/odilia-app/atspi/commit/fac1e86366f1e01e90a603401d7a90088e540583
cargo fmt # to trim down and reformat.
```
Only minor editing was done in `identify.rs`.

TTWNO:
Just checked out the latest changes. Wow! Loving it! This is basically end-to-end testing here.

These tests guarantee that:

1. The event exists in at-spi2-core.
2. The event is using the (mostly) right types, signatures on our side.

Which is some great stuff for keeping up with XML updates. I'm in your debt for this one!

luukvanderduim:
> Just checked out the latest changes. Wow! Loving it! This is basically end-to-end testing here.

Thanks!
  
>     2. The event is using the (mostly) right types, signatures on our side.

Well.. I have some things on my wish-list in that department. 
For instance, consider:
 
```rust
    impl PropertyChangeEvent {
        #[must_use]
        pub fn value(&self) -> &zbus::zvariant::Value<'_> {
            self.0.any_data()
        }
    }
```
As a user I would find this discouraging.
This works, but the end-user still is required to look-up the the intended type for the case.

I feel we should try to find a way to let the method evaluate to a more meaningful type to our users.

luukvanderduim:
To do:
- [ ] Move mock-up event to a test tree.
- [ ] Add tests for the 'reinstating comments' logic. Demonstrate reinstating is somewhat useful, otherwise sunset the whole idea.
- [ ] Reinstate tables and other useful comments.
- [ ] Merge with main again.

TTWNO:
If you edit the original comment opening the PR, you can have the task completion show up in the issues list. Obviously, you don't need to do this, but just so you know.

luukvanderduim:
> If you edit the original comment opening the PR, you can have the task completion show up in the issues list. Obviously, you don't need to do this, but just so you know.

You read my mind! I had indeed been wondering that.. Thank you!

TTWNO:
Looks like I can edit the comment too, haha!

TTWNO:
How are you emitting events with those tests? I can see they pass, but I don't see any code that actually emits the event?

The tests also test the limits of my machine (in terms of RAM usage), not sure if that's your tests, or some side-effect of trying to open a bunch of async runtimes all at once.

TTWNO:
Found the RAM problem. It's `ld` lol! I guess compiling and linking against the examples when `cargo test` is run can be quite intensive.

luukvanderduim:
> How are you emitting events with those tests? I can see they pass, but I don't see any code that actually emits the event?

They don't.. but hear me out..
The doctests demonstrate the `# Examples` is relevant and in sync with the library. Please take a look at the excellent: [Making Great Docs with Rustdoc](https://www.tangramvision.com/blog/making-great-docs-with-rustdoc#-examples) guide too.

 As effect of being generated it also guarantees it has an upstream event in XML to back it up.
These examples show how one would want to use it and the text points the user to the contextually more complete examples. The examples are deliberately concise, to offer insight at a glance and ensure easy maintainability.

That being said, I agree that the types need test coverage beyond this: 
I am working on generating proper Rust CI tests - similar to the one we have in `atspi/tests` : 

Doing it in in Rust CI tests is easier and cleaner because you can abstract away shared functionality where you cannot in doctests. For example, having a private helper function to emit the signal requires a helper crate to get this to work in doctests while in CI tests you can just add the helper to the `common` library functions.
Also writing and debugging CI tests is easier because rust-analyzer aids you when writing CI test code, where it cannot in doctests, where you get feedback at compile- and/or test time only.

We will get the coverage we  want, and more but not in doctests.

(I have tried for hours to get private helper functions for the examples to work only to find that doctests treat the library as external and as such one cannot have access to private helpers in the library. Here the Rust developers nudge us towards the CI tests)

I  have also spent a bit of time on devising some other functions to craft faulty (and vaid) `Message`s to test the way our crate handles those.. so stay tuned.

> The tests also test the limits of my machine (in terms of RAM usage), not sure if that's your tests, or some side-effect of trying to open a bunch of async runtimes all at once.

I can recommend [mold](https://github.com/rui314/mold). It is much faster than `ld` and is written by the author of LLVM ld. I usually just check out and build from source, but there are many packaged versions too (such as AUR). I just noticed it has a section on how to speed up GitHub Actions with mold too.

TTWNO:
Could you write some tests for the doc reinstating code?

Other than that, this looks great!

luukvanderduim:
> Could you write some tests for the doc reinstating code?
> 
> Other than that, this looks great!

There are 26 tests covering collecting and reinstating. You want more on reinstating?

TTWNO:
Oh I see! They're in `atspi-codegen`. You're all good! Let's get 'er merged!

TTWNO (approved):

TTWNO:
You have to approve since I made the latest change.

TTWNO:
Nvm, we chatted on Matrix.

Issue #37, opened 2022-12-29
Event de-structuring is cumbersome.


luukvanderduim:
Currently the event de-structuring looks as such:
```Rust 
while let Ok(ev) = event_stream.next().await.expect("None") {
    let Event::Interfaces(EventInterfaces::Object( ObjectEvents::StateChanged(state_ev))) = ev else { continue; };

    if state_ev.state() == "focused" {
        let Some(sender) = state_ev.sender()? else { continue; };
        println!("Focused by: {sender}")
     }
}
```
If `Event` would hold a copy of `AtspiEvent` (instead of `EventInterfaces`) and the `TryFrom` implementations are adjusted to convert from `Event` to respectively: the signified types or the grouped `EventInterfaces`, we could:
 
```Rust
while let Ok(ev) = event_stream.next().await.expect("None") {
    let ev: StateChangedEvent = ev.try_into()? else { continue; };

    if ev.state() == "focused" {
        let Some(sender) = state_ev.sender()? else { continue; };
        println!("Focused by: {sender}")
     }
}
```
perhaps have current behaviour as such:

```Rust
while let Ok(ev) = event_stream.next().await.expect("None") {
    let evs: EventInterfaces = ev.try_into()? else { continue; };

    /// Then do destructuring ...
}
```

TTWNO:
It may be cumbersome, but it is very easily typed and has a good structure.

If you want a way to do this with AtspiEvent, then why not with Cache and Availsble events too? And at that point, what's the use of the enum?

I think a good comprimise would be to have *all* event variant in Event itself, without splitting them into individual, progressively more specific, enums. Either work well for Odilia.

I'll ping @DataTriny and see what they think.

TTWNO:
Resolved by #50

PR #50, opened 2023-01-15
 Destructure Event without need to know its structure.

luukvanderduim:
Addresses #37 

Atspi separates events by bus-type signature and groups them accordingly in enum structures, which in the end are grouped by the `Event` enum.
This event tree allows `atspi` to emit a conveniently typed heterogeneous variety of events using the same stream `event::event_stream()`.

To some users, the way we structured these events may be of little interest and may obscure or muddle their goal of obtaining the final type.
This PR allows users to hide the nested enum structure altogether. The only thing they need to know is the event of interest.

The `let-else` allows users to bind by naming the type and calling its `try_from` implementation on the nested event,
 which succeeds or continues to the next event in the case of a mismatch.

```rust
    while let Some(Ok(ev)) = stream.next().await {
        let Ok(change)  = <StateChangedEvent>::try_from(ev) else { continue };

	// Do interesting things with `change` the `StateChangedEvent`.
    }
```
A follow-up commit in this PR adds two examples using the tokio and the async-std executors to demonstrate a basic `atspi` usage.

```shell
cargo run --example focused-async-std
# or
cargo run --example focused-tokio
```


TTWNO (commented):
This update should be done through atspi-codegen. It's otherwise quite sound.

TTWNO:
There is currently no way to de-structure Available, Cache, or Listener sub-enums. Only the `EventInterface::*` type. Is this ok, or do we want it for all subtypes?

If so, we should move more code to auto-generation.

luukvanderduim:
> There is currently no way to de-structure Available, Cache, or Listener sub-enums.

There was, but you deleted them: 

```rust
#[rustfmt::skip]
impl TryFrom<Event> for AddAccessibleEvent {
    type Error = AtspiError;
    fn try_from(ev: Event) -> Result<Self, Self::Error> {
        if let Event::Cache(CacheEvents::Add(event)) = ev {
            Ok(event)
        } else {
            Err(AtspiError::Conversion("invalid type"))
        }
    }
}

#[rustfmt::skip]
impl TryFrom<Event> for RemoveAccessibleEvent {
    type Error = AtspiError;
    fn try_from(ev: Event) -> Result<Self, Self::Error> {
        if let Event::Cache(CacheEvents::Remove(event)) = ev {
            Ok(event)
        } else {
            Err(AtspiError::Conversion("invalid type"))
        }
    }
}

#[rustfmt::skip]
impl TryFrom<Event> for AvailableEvent {
    type Error = AtspiError;
    fn try_from(ev: Event) -> Result<Self, Self::Error> {
        if let Event::Available(event) = ev {
            Ok(event)
        } else {
            Err(AtspiError::Conversion("invalid type"))
        }
    }
}

#[rustfmt::skip]
impl TryFrom<Event> for EventListenerRegisteredEvent {
    type Error = AtspiError;
    fn try_from(ev: Event) -> Result<Self, Self::Error> {
        if let Event::Listener(EventListenerEvents::Registered(event)) = ev {
            Ok(event)
        } else {
            Err(AtspiError::Conversion("invalid type"))
        }
    }
}

#[rustfmt::skip]
impl TryFrom<Event> for EventListenerDeregisteredEvent {
    type Error = AtspiError;
    fn try_from(ev: Event) -> Result<Self, Self::Error> {
        if let Event::Listener(EventListenerEvents::Deregistered(event)) = ev {
            Ok(event)
        } else {
            Err(AtspiError::Conversion("invalid type"))
        }
    }
}

```
I will reinstate these in `event/mod.rs` so they will accompany their respective events. I am not too worried about these conversions becoming invalidated by changes in type or enum structure because, if those change, the conversions will fail to build. We would notice before harm is done.

#### code generation 

We should expect upstream to revise signatures and a few deprecations but I expect event types to stay the same, but there might be exceptions.  Their methods might change (signatures). Thinking of `IsEnabled` again. 
These conversions however will probably go mostly untouched, unless we decide to refactor enums.  

When changes arrive, we want to swap the generated interfaces easily. Then regenerate identified events and impls and restore docs.

Should we physically reorder such that it is clear to us (or future maintainers) what not to touch because it is regenerated?
eg. having a folder called `auto` or something similarly named?

TTWNO:
> There was, but you deleted them:

Whoopsie! Good catch!

TTWNO:
> Should we physically reorder such that it is clear to us (or future maintainers) what not to touch because it is regenerated?

Having a comment that the file is auto-generated should be sufficient.

TTWNO:
Some "manual" auto-generation techniques needed to be used here because of how we name our structs, etc.

This should be able to go away once we have all of that auto-generated as well. Thoughts on this method?

My main concern is that new signals will come along which will not implement these, and I don't want to forget.
Since Rust doesn't have parameterized tests at the moment, we can't exhaustively test all sub, and sub-sub-enums, etc. for a working implementation of this trait.

So I've switched this one over to use auto-generation with some hacks in there for now. With the long-term goal to be to remove them and use nearly-fully auto-generated code.

luukvanderduim (commented):
What else needs to be done here?

TTWNO (approved):

TTWNO:
I think that's it. You can now destructute events without all the outer types.

Issue #49, opened 2023-01-11
AccessibleExt should be cache-transparent


TTWNO:
AccessibleExt, and functions like it should contain methods that work on cache or non-cache-enabled versions of `atspi`.
For example, Odilia should be able to use its own caching system to get accessible elements further along in a document, so there should be some way for us to either get the related AccessibleProxy, *or* convert from a caching function/type that can be converted into one.

Then methods like `get_role()` could be run in a cache-transparent manner.

The exact way to do this is unknown, since `Into<AccessibleProxy>` won't work (async call required to create it), and even if we could build one, the AccessibleProxy will not be able to have access to a cache.

Any ideas? We could implement a custom trait like GetFromCache for AccessibleProxy and go from there, but I think there has got to be a better solution.

Speaking of better solutions, @DataTriny , hate to ping you, but how does AccessKit handle its cache? Is there any way that we could have a trait which would give you access to all AccessibleExt functions (and future additions) without the need to directly write your own version of every function?

I'm thinking something like this:

```rust
pub trait AccessibleMaybeCache {
    pub async fn parent() -> impl Accessible;
    pub async fn children() -> Iterator<Item = impl Accessible>;
    pub async fn application() -> impl Accessible;
}
```

I don't think these methods would *need* to be async, so perhaps even use something like `maybe_async`? I'm throwing ideas to the wall here and hoping something sticks. I don't want to force async onto the caching implementation if it works without it.

This way, something like `get_next` could look like: `AccessibleExt::get_next<T: AccessibleMaybeCache>(root: &T, matcher_args: &MatcherArgs, backword: bool);`

Does AccessKit have a function like this already? If so, would it be realistic for it to work with this atspi caching-transparent method, or would you rather just handle it all on your own. I realize AccessKit is quite a larger project than Odilia so I would understand wanting more fine-grained control over your end-product.

(Sorry about the stream-of-consciousness issue here. I'll fix it up later.)

luukvanderduim:
> AccessibleExt, and functions like it should contain methods that work on cache or non-cache-enabled versions of `atspi`. For example, Odilia should be able to use its own caching system to get accessible elements further along in a document, so there should be some way for us to either get the related AccessibleProxy, _or_ convert from a caching function/type that can be converted into one.

So you have these in a `CacheItem`:

```rust
    /// The accessible object (within the application)   (so)
    pub object: (String, OwnedObjectPath),
    /// The application (root object(?)    (so)
[...]

PR #52, opened 2023-01-22
Foreign impersonations of accessibles

TTWNO:
This PR will merge support for providing your own `Accessible`s to the rest of the `atspi` crate, like `AccessibleExt`.

By the end of this refactor, all `*Ext` methods will be implemented like so:

```rust
trait AccessibleExt { ... }
impl<T: Accessible> AccessibleExt for T {
   ...
}
```

This includes convertable, which will check *only* for the `T: Accessible` bound. All current traits like `ProxyDefault` and `AtspiProxy` will need to be auto-generated and related parts implemented by the `Accessible` teait.

This should be painless once set up through macros. Current progress is as follows:

* [X] Pass through parameters to `dbus_proxy` macro.
* [X] Auto-generate traits for blocking and non-blocking versions of the proxy.
* [x] Provide default implementations for `*Proxy` and `*ProxyBlocking` structs.
* [x] Convert all existing `*Ext` traits to be implemented for `*` so `AccessibleExt` would be implemented on `Accessible` and `TextExt` on the `Text` trait. (Only done for non-blocking proxies).
* [x] There should be automatic implementations of `*ExtError` for all existing `*Ext` modules.
* [x] There should be an `*ExtError` and `*Ext` for every interface in the `atspi` crate; this will make it more forward-compatible when the time comes to implement additional extensions to existing interfaces. Although no more strike me at this time, I'm sure this will happen sooner rather than later.

Hoping this can be wrapped up by next month, but this is a fairly large issue which I expect will take a while.


TTWNO:
Now I need to change the Convertable trait to be able to work with these traits.

I think I can get this to work, but I may need to add more methods to the trait, or create a second trait.
Right now, I'm depending on the AtspiProxy, ProxyDefault and From<Proxy<'c>> traits to create an extremely generic `Convertable` trait which has worked really well so far. The issue is that converting it to a new set of traits is going to be annoying since I don't have equiv.s with my traits.

For example: `From<Proxy<'c>>` gives us assurance that a ProxyBuilder exists for that struct.
We're using `ProxyDefault` to set the interface of the created Proxy struct.
We're using `AtspiProxy` so that when we query a proxy's atspi interfaces, we can return an error in the case that the requested proxy's interface doesn't match the list of possibilities.

Not sure how to go about refactoring all this. It's very delicate lol!

TTWNO:
Everything is working, despite a manual implementation of `AccessibleExtError` and `TextExtError` for both AccessibleProxy and TextProxy respectively. This should be automated, or at least auto-generated as a one-of like with `identify.rs`.

This should cause no breaking changes, except that the `get_text_ext` method has been renamed to `get_all_text`.

In addition, all of these `*Ext` traits should probably be async and non-async. We may need the `maybe_async` crate, but I don't know how much of a pain that will be to set up.

Anyway, here's the rundown of what we have at this point:

1. If you implement `Accessible`, `Convertable`, `Send`, `Senc`, and `AccessibleExtError` on any struct, you should automatically get all methods in the `AccessibleExt` trait.
2. You need to implement an error type that satisfies various trait bounds to be able to move to get all those methods for free, but at least you can now.
3. Most types in `Accessible` and `AccessibleExt` have become generic: `Self`. Meaning that you must return the same type as you started with to satisfy trait bounds. So for example, if you want to implement `parent()` for your custom struct, you must return something which has the same type as your original struct.
4. If you implement `Text`, and `TextExtError` on any struct, you should autoamtically get all methods in the `TextExt` trait for free.

I want to genericize this further in three ways, somewhat stated earlier:

1. There should be a non-async option for those who would like to use blocking code.
2. There should be automatic implementations of `*ExtError` for all existing `*Ext` modules.
3. There should be an `*ExtError` and `*Ext` for every interface in the `atspi` crate; this will make it more forward-compatible when the time comes to implement additional extensions to existing interfaces. Although no more strike me at this time, I'm sure this will happen sooner rather than later.

Finally, there needs to be a lot more documentation; this is overwhelming for a newbie, especially when they're new to Rust's type/trait system as is.
I've edited the original post to reflect these desires. But big progress is being made.

TTWNO:
TODO: create non-async version of `AccessibleExt` and `TextExt`

TTWNO:
At this point, this should only require some cleanup.

`*Proxy` and `*ProxyBlocking` now implement `*(Blocking)Ext`, and as a consequence `*(Blocking)ExtError`. This has been done by a custom, `atspi_proxy` macro to replace to `dbus_proxy` macro. Note that the dbus_rpxoy` macro is still run, just from `atspi_proxy` and then additional code gets generated on top.

I have also added tests to ensure that the default `*Proxy` and `*ProxyBlocking` always implement their respective `*(Blocking)Ext` traits.

I'm very happy with this solution, and it also shouldn't break any existing code, since `AccessibleExt` was the only major `async_trait` that was used, and it is *still* indeed implemented for `AccessibleProxy(Blocking)`. The only difference is that now other structs may also implement traits like `*(Blocking)Ext` and `*(Blocking)ExtError` to automatically receive a`*Ext` trait implementation.

This is awesome for our caching system, because it means any code written to work with proxies: blocking or not, will now be able to be used on `CacheItem`s granted the impement the right traits.

Going to be at fossdem this weekend, so I'm not expecting any code updates until Tuesday.

TTWNO:
What the hell? Ok, I can't figure this one out.

TTWNO:
Found it: `zbus` -> `async_boardcast` -> `parking_lot` -> `parking_lot_core` -> `windows-sys v0.45`, but `tokio` -> `windows-sys v0.42`.

I can't really submit an upstream bug because this is just a mismatched requirement of a version.

This isn't going to be possible to fix realistically, so I'm disabling this error in `lib.rs`.

TTWNO:
The path is irreverent on any `(so)` types, since we only need the ID, I'm changing it to only store the ID as the `AccessibleId` enum.
The `sender`/`destination` string is still arbitrary I'm pretty sure so I'm leaving it for now.

TTWNO:
I'm going to try to not touch this for now. I want input from @DataTriny .

I can confirm that AccessKit compiles and passes all tests on this branch in its current sttate. Please let me know if you would like to see any changed before I merge this.

Also paging @luukvanderduim to do a code review of this, when you have time. This is in it's nearly finished state.
I suspect the `atspi-macros` code will need some tuning, refactoring.

TTWNO:
This issues fixes https://github.com/odilia-app/atspi/issues/49

atspi-macros/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "atspi-macros"
 version = "0.1.4"
-edition = "2021"
+edition = "2018"
```
luukvanderduim, line 4: Why downgrade the edition?

atspi-macros/src/utils.rs
```diff
@@ -0,0 +1,242 @@
...
+pub fn pat_ident(pat: &PatType) -> Option<&Ident> {
+    match &*pat.pat {
+        Pat::Ident(PatIdent { ident, .. }) => Some(ident),
+        _ => None,
+    }
+}
+
+// Convert to pascal case, assuming snake case.
+// If `s` is already in pascal case, should yield the same result.
+pub fn pascal_case(s: &str) -> String {
```
luukvanderduim, line 35: Nitpick: [UpperCamelCase](https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md) is more common in Rust than PascalCase.

atspi-macros/Cargo.toml
TTWNO: Ah I fixed the reason I had to change that in the first place.

TTWNO:
This shouldn't affect `accesskit`, @luukvanderduim do you think this is ready for merge?

Let me know, and I'll merge this once I fix the inevitable conflicts with the other major branch landing soon.

DataTriny:
Hello,

Sorry I am busy with other things at the moment. What is the impact on dependency count and binary size of this PR? This is our top priority for AccessKit right now and I can see a few changes to `Cargo.toml`.

TTWNO:
Static assertions only adds a frw macros and has no dependencies of its own.

Anything in dev dependencies can safely be ignored for binary size/compile time as it's only used for tests and docs.

TTWNO:
CC: @DataTriny, once this PR is merged, you will want to specify these features:

```
atspi = { default_features = false, features = ["tokio"] }
# or
atspi = { default_features = false, features = ["async-io"] }
```

The default features now contain a few additional dependencies to make client implementations easier. This PR, which re-introduces many dependencies in `atspi-macros` has those additional dependencies now feature-gated as well behind the `unstable-traits` feature.
This will also drop a half-dozen additional dependencies for accesskit.

TTWNO:
I'm merging this ASAP because it is diverging too far from `main` to be maintainable. There are multiple pain points with merging already and I don't want that to continue. Incremental improvements will be made, but it will still be feature-gated behind an `unstable-*` feature until it's ready, then the `unstable-` part of the flag will be removed once it's stabilized.

luukvanderduim:
> I'm merging this ASAP because it is diverging too far from `main` to be maintainable. There are multiple pain points with merging already and I don't want that to continue. Incremental improvements will be made, but it will still be feature-gated behind an `unstable-*` feature until it's ready, then the `unstable-` part of the flag will be removed once it's stabilized.

Big changes get increasingly difficult to merge, lets merge.
I have ideas to improve upon this but also too many unknowns to act.

luukvanderduim (approved):
A gated unstable feature and improve upon it from there.
May well be preferable over a divergent tree, which is painful.
Lets see how this goes.

PR #53, opened 2023-02-08
Make registering for events easy

### LINKED ISSUE #34: Make registering for events easy
Right now  one has to use bpth atspi and zbus separately to allow events to come in on an event stream.

It should be this easy:

```rust
let conn = connection::new();
conn.register_event(Event::Document::LoadComplete);

let stream = conn.event_stream();
while let Some(event) = stream.next().await {
    ...
}
```

If the library user would like to add additional things to the dbus connection, then they are welcome to do so by getting the internal connection (already possible) or creating a new connection all on their own.

It should be easy to use `atspi` without knowing how dbus works—you should only have to know that this is hpw atspi works, and not be exposed to implementation details.

TTWNO:
Fixes #34


atspi-codegen/src/bin/event_names.rs
```diff
@@ -0,0 +1,116 @@
...
+	let iface_data = data.interfaces()
+		.iter()
+		.map(|iface| generate_mod_from_iface(iface))
+		.collect::<Vec<String>>()
+		.join("\n");
+  format!("{iface_data}")
+}
+
+pub fn main() {
+	println!("/// BEWARE: This file has been automatically generated by the `atspi-codegen` crate.
```
luukvanderduim, line 105: Functions in binaries are never exported. Why is main 'pub'?

src/events/names.rs
```diff
@@ -0,0 +1,140 @@
+/// BEWARE: This file has been automatically generated by the `atspi-codegen` crate.
+/// Any edits to this file will be overwritten by the next re-generation of this file.
+
+use serde::{Serialize, Deserialize};
+use enumflags2::bitflags;
+
+pub trait Signal {
+  const INTERFACE: &'static str;
+}
+#[bitflags]
```
luukvanderduim, line 9: We likely want to annotate this enum with  #[non_exhaustive]. If it is even remotely possible the enum to change during its life, it should be. Adding #[non_exhaustive] allows us to add/remove variants while staying backwards compatible and  without breaking other peoples code.

See for reference: [Prevent breaking changes with non_exhaustive](https://blog.knoldus.com/prevent-breaking-code-changes-in-future-releases-using-non-exhaustive-enums-in-rust/)

luukvanderduim: (Of course the same goes for all other public enums here.)

luukvanderduim (commented):
Did you intend this filename 'deserializie.rs'?

Can this test fail?

luukvanderduim (changes requested):
Why generate new enums at all? We already have event enumerations.

If I can inspire you to a change of mind on this feature: consider reusing already existing public enums and types.
That will reduce the code considerably, reduce exposed types considerably and it prevents repeating the same (kind of) enums.

If the existing events and enums gain type level information on their interface, we could do something like this short-hand match-rule too:

```rust
use identify::StateChangedEvent;

// Match rule acquiring through the existing type:
let rule = StateChangedEvent.rule()?; 
```
This `rule()` method may not always cut it as, in some cases, one could want to specify eg. `ButtonEvent.rule()` which button, and if it is either released or pressed. 
But we could argue that if one wants to be that specific, they would not use the shorthand  but rather the verbose rule builder.

Curious to hear what you think.

src/connection.rs
```diff
@@ -155,6 +154,62 @@ impl Connection {
...
+        };
+		let match_rule = MatchRule::builder()
+			.msg_type(MessageType::Signal)
+			.interface(<T as Signal>::INTERFACE)?
+			.member(member_string)?
+			.build();
+		dbus_proxy.add_match_rule(match_rule).await?;
+		Ok(())
+	}
+
```
luukvanderduim, line 188: I am not sure we want to hide the MatchRules like this.
This combines registering with registryd and adding a MatchRule to the bus, dbus-broker.
Two different daemons and separate connections.

luukvanderduim (commented):
Adding type level information on the events seems like a very good idea. This may allow some operations to even move to compile time.

I wonder if we really need to introduce all these new enums. We should try to reuse those we already have. Having similarly named enums that do only slightly different thins is confusing and adds to exposed surface.

Hiding MatchRule registering completely may look convenient but may lea to surprising behavior and that goes against the principle of 'least surprise'.

src/connection.rs
```diff
@@ -155,6 +154,62 @@ impl Connection {
...
+	/// Register multiple events in one swoop!
+	/// Very useful for registering events of one interface together.
+	/// This can be done like so:
+	///
+	/// ```rust
+	/// use atspi::events::names::ObjectEvents;
+	/// use enumflags2::BitFlag;
+	/// # tokio_test::block_on(async {
+	/// let connection = atspi::Connection::open().await.unwrap();
+	/// connection.register_events(ObjectEvents::all()).await.unwrap();
```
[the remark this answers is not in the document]
TTWNO: Not sure where you're proposing I use `any()`? That isn't a function provided by `enumflags2`.

tests/deserialzie.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(test)]
+mod tests {
+	use atspi::events::names::ObjectEvents;
+	use serde_plain;
+
+	#[test]
+	fn serialize_object_variant() {
+		assert_eq!(
```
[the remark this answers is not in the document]
TTWNO: We use it to set up the match rule, see `src/connection.rs` line 177. Of course, we could also set up an infallible `ToString` trait as well. That would probably be better since that reduces our dependencies.

src/connection.rs
TTWNO: This is incorrect on two counts:

1. This *only* adds a match rule, it does not add anything to the registry. The registry doesn't seem to do anything. I have tested this pretty extensively, and I see nothing additional coming down the bus based on `registryd`; this seems to only by for key bindings at this point.
2. It uses the same connection; check `odilia/src/main.rs` lines 35-40.

This *does* hide match rules; but, of course, it does not stop a user from manually creating a DBusProxy, and adding a `MatchRule` manually. This is always an option. It seems to me though that is is unnecessary. The MatchRule can be safely hidden from the programmer.

It would also be fine if those types from `identify.rs` would have a static `Into<MatchRule>` implementation. Something like `identify::object::StateChangedEvent::into::<MatchRule>()`. This would allow the user of the library to know that the matchrule exists, and be able to modify it without needing to create it themselves.

Thoughts? I'd be fine with an implementation similar to that. I'll get working on it asap.

luukvanderduim: >     1. This _only_ adds a match rule, it does not add anything to the registry. The registry doesn't seem to do anything.

[This is what I could find it does:](https://web.archive.org/web/20140414005650/http://www.linuxfoundation.org/collaborate/workgroups/accessibility/atk/at-spi/at-spi_on_d-bus#AT-SPI_D-Bus_Specification)

"
The registry daemon has two distinct tasks.

    Keeping track of all the accessible applications on the desktop.
    Dealing with and forwarding X events over AT-SPI D-Bus.
"

luukvanderduim: > This is incorrect on two counts:

No more reviews after midnight for me.

> 
>     1. This _only_ adds a match rule, it does not add anything to the registry. 

Ah, yes. Right. I think the name made me think it did registering too.

> 
> This _does_ hide match rules; but, of course, it does not stop a user from manually creating a DBusProxy, and adding a `MatchRule` manually. This is always an option. It seems to me though that is is unnecessary. The MatchRule can be safely hidden from the programmer.

Again I think I jumped to the conclusion that it registered events because of its name.
It adds a MatchRule instead.

> 
> It would also be fine if those types from `identify.rs` would have a static `Into<MatchRule>` implementation. Something like `identify::object::StateChangedEvent::into::<MatchRule>()`. This would allow the user of the library to know that the matchrule exists, and be able to modify it without needing to create it themselves.

Sounds really good.

> 
> Thoughts? I'd be fine with an implementation similar to that. I'll get working on it asap.

You know we get `Into<MatchRule> for StateChangeEvent` for free if you implement `From<StateChangedEvent> for MatchRule`?

TTWNO: Unfortunately, From seems to take a tegerence to self, which would require an instanciation of the type. In this case that is not necessary, but the MatchRuleBuilder functions aren't const so I'm just building an adsociated function to run at runtime.

The latest version of this PR implements a `match_rule()` and `match_rules()` associsted function for the struct snd enum containing it respectively. Check that out and let me know.

luukvanderduim (approved):

luukvanderduim (approved):

luukvanderduim (approved):

luukvanderduim (approved):

src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![deny(clippy::all, clippy::pedantic, clippy::cargo, unsafe_code)]
 // #![deny(clippy::missing_docs)]
+#![allow(clippy::multiple_crate_versions)]
```
luukvanderduim, line 3: I believe this is fixed in upstream `zbus` as I no longer see these warnings where I did february 10th.

luukvanderduim (approved):

luukvanderduim:
Should we change `Connection` to contain both `DBusProxy` and `RegistryProxy` and add an impl with `as_dbus_proxy()`, `as_registry_proxy()` and an assortment of methods of most used methods on either Proxies in this branch?

If we do, can I suggest to rename that struct to `A11yBus` because that is what it represents and avoids unneeded confusion with `zbus::Connerction`?

If you want me to tackle these here or in a follow up, that is fine too of course!

TTWNO:
We can do that. If you have time, great! If not, I'll aim for the end of the week.

src/lib.rs
TTWNO: zbus v3.10.0 seems to not have solved this problem for me. I'm on latest stable and ran `cargo update` before checking with `cargo clippy`.

TTWNO:
I found a way to use `const`s instead of function calls, with the downside that it needs to convert to a MatchRule on the fly.

Just need to implement the de-register methods, then we're good.

TTWNO:
Ahhh! The way to call these is so silly, I'm going to pass the `const`s instead of using generics.

luukvanderduim:
> EDIT: but that would allow random strings to be added to the registry/matching rules potentially.... Grrr.

Not sure if I understand the problem fully, but would 'lazy_static' solve that? 

> I found a way to use `const`s instead of function calls, with the downside that it needs to convert to a MatchRule on the fly.

I like it, however, note that users typically build Match Rules only once-ish?

TTWNO:
Generics are fine, I suppose. At least it allows us to generically register events for any `enum` *type*, and not the variants.

I'm happy with this PR. Unless you have anything to add, let's merge this.

luukvanderduim (approved):

PR #54, opened 2023-02-09
Add Semver Cheks to CI

TTWNO:
This will allow us to know (more-or-less) if a breaking change has been made. For now, we'll use this as a warning to update the version number, but in the future this could be helpful to stop PRs from accidentally modifying the public API.


TTWNO:
I tried doing this on main for a while; probably close to 10 commits. RIP my email inbox.

PR #58, opened 2023-02-19
Remove abandoned XML-based code generation from atspi-macros

### LINKED ISSUE #57: Consider dropping the apparently-unused XML-based generation from atspi-macros
Unless I'm missing something, it appears that the `create_from_xml` macro in atspi-macros isn't being used, in either atspi or Odilia. If that's correct, then I strongly recommend dropping that macro from this crate. Then, this crate won't need to depend on zbus and related crates. Eliminating zbus from the build-time dependencies will have a major impact on compile time, especially in release builds. If the plan was to start using `create_from_xml` later, please reconsider that plan. The number of dependencies introduced by atspi and zbus is already causing some friction in a current AccessKit integration project; I'd like to get those numbers down.

If there's no objection, I can open a PR to make the requested change.

mwcampbell:
fixes #57


TTWNO (approved):

TTWNO:
Hate to be a pain, but could you bring in the latest commit from main. I want futute contrinutors to have working GH Actions, so just making sure it works wirh your fork here.

Thanks!

mwcampbell:
No problem, just rebased

TTWNO (approved):

DataTriny:
Woah, AccessKit's winit simple example shrank by more than 1MB on Linux just by using the latest code on `main`.

TTWNO:
Yes. I beleive the version accesskit uses is a bit behind. We re-worked the convertable.rs file to be about 1/20th the size that it was previously.

mwcampbell:
AccessKit is using atspi 0.8.7, which I think was released after you reworked convertable.rs.

To my great surprise, the impressive reduction in binary size (when optimizing for speed) is due entirely to this PR. I didn't think this PR would reduce binary size, only build time. What appears to have happened is that since atspi-macros enabled the xml feature in zbus, that feature was also enabled in the runtime build of zbus, even though it was only wanted at build time (and not really even there, which was the point of this PR). For whatever reason, the difference in binary size is much less dramatic when optimizing for size (`opt-level = "z"`); in that case, it's only about a 40K difference. Obviously when optimizing for speed, the heavy inlining magnifies the impact of including XML serialization support in zbus.

TTWNO:
Great investigation! Ongrats to accesskit!

mwcampbell:
Can I please ask for new published versions of atspi and atspi-macros, when you have a chance, so we can update AccessKit? Thanks.

TTWNO:
All done! `v0.10.0`

As always, let me know if there's any other way I could help.

PR #66, opened 2023-03-10
Add sparse-registry option to all builds

TTWNO:


TTWNO:
Looks like it shaves about a minute off of a few of our CI pipelines. Seems worth it to me!

TTWNO:
Semver is failing because of a version mismatch problem. This is only because of git tagging, so we can ignore it for this.

luukvanderduim:
As this only changes CI, this does not impact the minimal Rust compiler for users, or does it?

I am side-eyeing stabilized `pin!` from std.

TTWNO:
> As this only changes CI, this does not impact the minimal Rust compiler for users, or does it?

Correct.

> I am side-eyeing stabilized pin! from std.

I think that only really allows us to use it instead of `tokio::pin!()`, so it wouldn't really help depedencies. But maybe we could disbale some features? Anyway, different issue. I'll open a new PR/issue for it.

PR #67, opened 2023-03-10
Remove `tokio::pin!` usage in favor of `std::pin::pin!` (and add integration tests to every event)

TTWNO:


TTWNO:
This has removed `tokio::pin!` from our crate, and adds end-to-end tests for every event in `atspi` :)

Going to try removing some extra `tokio` features now.

TTWNO:
We should probably try to make that call out to `busctl` a macro. We could create a `test` feature and put it behind that feature flag. `atspi-test` would absolutely be a valid use case considering there's a lot of boilerplate.

TTWNO:
Also, note that we do not need to update MSRV, since this is only a test dependency. Not sure if there's a way to specify MSRV for tests.

PR #76, opened 2023-04-26
Add mold to CI, fix failing tests

TTWNO:


TTWNO:
What the hell? Why does this test work now? I didn't even actually fix it; I just tried to log the error more accurately.

luukvanderduim:
> What the hell? Why does this test work now? I didn't even actually fix it; I just tried to log the error more accurately.

I think I understand now the flakiness the async tests exert. 
What happens is that the test expects that the first event from the timeout stream is the event we are looking for, but that is not necessarily true. In another test I got a DBus event instead of the `Cache` event we expected the test failed due to the expectation.

The fix would be to handle events from the stream until the event of interest is received or until timeout.
Something along these lines:

```Rust
while let Some(event) = to.await.next() {
    match event {
        Ok(Event::Cache(CacheEvents::Add(event))) => { ... // asserttions },
        Ok(_some_other_event) => { continue; }  // we dont care about other events, lets see the next!
        Err(_) => { panic!("timed out") }
    }
}
```
The reason why it sometimes worked and sometimes wouldn't is because, even in the CI environment, other things happen on the bus.

PR #78, opened 2023-04-30
Simplify Event Types

TTWNO:
Change from a model of using methods on top of raw `zbus::Message` data, to using plain fields.
The event is destructured (currently somewhat violently, with `.unwrap()`) into any of the corresponding event types.

This frees `Event*`/`*Event` from their underlying `body` and `message` fields that made creating new events rather difficult.

Since a `TryFrom<Arc<Message>>` already exists all event types, you can always use `.try_into()` to get an event back into a message.

This makes using the event types much easier and more separated from the underlying `zbus` type.

EDIT: It's late here, so not sure how much sense this makes. I'll read it over in the morning and make needed corrections.

TODO:

* [x] Remove unwraps
* [x] Implement `GenericEvent` and `TryFrom<Arc<Message>>` for all (struxt and enum} event types.
* [x] `Default` implementations for all (struct) event types.


TTWNO:
Damn, those tests were confusing me. Finally got it. Implemented a few more features than I was expected for one PR, but it can be summarized below:

1. Remove `Identify` trait.
2. Remove `Signify` trait.
3. Flatten `InterfaceEvents` enum into `Event` directly.
4. Remove the `newtype` pattern for all events. Replaced with conversion functions. All events now only contain the fields required to build them. You can now construct a TextCaretMovedEvent like so: `TextCaretMoved { item: Accessible { name: ":0.0".try_into()?, path: "/org/a11y/atspi/accessible/null".try_into()?, }, position: 22, properties: HashMap::new() };`. It's a bit verbose right now (because I think properties isn't necessary), but this makes our tests MUCH simpler, and allows `atspi` to function as both a sender and receiver.
5. All events now implement `From<&zbus::Message>` and `zbus::Message` now implements `TryFrom<ALL_EVENTS>` (via the `GenericEvent` trait, see below). You won't always want to send a message generated from this conversion, however, since sending an event with a UnqiueName you don't already own will cause DBus to send a "NameAquired" signal right back to you, so there is a separate functionality for testing, or, more accurately, for sending events using the already squired bus name, regardless of what is stored in the event, this is `AccessibilityBus::send_event()`.
6. New `GenricEvent` trait that defines a bunch of constants and common methods to access fields of structs so that `impl GenericEvent` can now be accepted into functions for handling any event.

Ok, I think that's it. This makes my test look the way I wanted them.

I'd love some feedback now that it's mostly done. Especially from @luukvanderduim and @DataTriny , maybe @mcb2003 if you have time.

I will probably do one more large PR before we hit `1.0`; I want borrowed types when possible, but right now we have a ton of `Owned*` or `*Owned`, or `ObjectPath<'static>`. Having all events be able to be created and converted without cloning fields would be good, when I get around to it.

TTWNO:
Closing in favour of #82

PR #80, opened 2023-05-04
Wasm Types

TTWNO:
Implement most types via `*.wai` translations so that the types can be passed over wasm FFI. This should not, under any circumstances, break compatibility from `v0.16.0`. Note that this branch depends on the [`simplify_events`](https://github.com/odilia-app/atspi/pull/78) PR getting merged.

TODO:

* [X] Convert `atspi` to a workspace.
* [X] Create new `atspi-types` crate.
* [ ] Move most types to the types crate, and generate them from `*.wai` bindings (`*Proxy` excepted).
* [x] `Event` (enum)
* [x] `*Event` (struct)
* [x] `*Events` (enum)
* [x] Generate `Default` implementation for `*Events` bindings
* [x] `Role` (bitflags)
* [x] `States` (bitflags)
* [ ] `RelationType` (enum)
* [x] `Accessible` (struct)
* [ ] Find a way to create the types *without* the `*.wai` bindings. We do not want to make it mandatory. Use feature flags.
* [ ] Re-export `atspi_types::*`
* [ ] Add compilation test to CI to ensure type consistency, even without wasm bindings.


TTWNO:
Closing, as I'm not satisfied with the state of the WASM ecosystem at this time.

PR #81, opened 2023-05-05
Split Types into subcrate; convert to workspace

TTWNO:
This is for the future so that the types atspi use will be able to be used on less supported targets, over FFI, network, socket, etc.

In particular, if Odilia is to implement addons over WASM/Lua, we will need to compile types separately from the main `atspi` crate as `zbus` takes a hard dependency on an async runtime.


TTWNO:
completed on #82

PR #82, opened 2023-05-11
Types crate v2

TTWNO:
* [X] Separate `-types` into their own crate. This crate should be compileable on basically any system.
* [X] Separate `-client` into its own crate. This crate is the equivalent of the old "client-accessories" features.
* [x] Fix all clippy warnings, add `--workspace` flag.
* [x] Fix semver checks (it only work on a crate-by-crate basis)
* [ ] Approval from other involved individuals.


DataTriny:
@TTWNO  I think it's very nice to separate the types in this way. This will probably speed up AccessKit's compile time a bit.

I'll review this, but it will take a while with all these changes.

TTWNO:
Semver check is broken due to https://github.com/obi1kenobi/cargo-semver-checks/pull/426

TTWNO:
Hi, @DataTriny . I hate to be a pain, but I have a lot of other PRs and future PRs based on these changes. Could you do a review soon?

The basics are this:

1. I've moved all types except the proxies in `atspi-types`.
2. Client connections, and convenience functions like `set_session_accessibility` have been moved into `atspi-client`.
3. Effectively, `atspi` has become `atspi-proxies`. I'm willing to rename this if you see fit.

All the tests will pass soon (just waiting on a new release of `semver-checks`), and we now cover over 20% more code than before. The way this is structured it should make everything much easier to test in the future.

Let me know how else I can make the review smoother for you.

DataTriny:
@TTWNO Sorry for being slow on this.

Preliminary questions:
- Is there a real benefit to the atspi-client crate? If it's purely a client utility, it would fit into atspi, if it's generic enough for both clients and providers then it could go into atspi-types.
- Why are the examples into atspi-types? Shouldn't they be better living inside atspi?

My understanding of your goal with this PR was to put everything common for clients and providers inside atspi-types, is this right? If yes, then atspi-types might not be the best name, maybe atspi-common or atspi-cor would be better?

TTWNO:
> Is there a real benefit to the atspi-client crate? If it's purely a client utility, it would fit into atspi, if it's generic enough for both clients and providers then it could go into atspi-types.

Good observation. It does indeed include a way to both send and receive events. I can put it behind a festure flag in the common crate? I'm also trying to have the common crate compile to WASM (which zbus doesn't support).

> Why are the examples into atspi-types? Shouldn't they be better living inside atspi?

Yep. I flubbed this one too. Good catch!

> My understanding of your goal with this PR was to put everything common for clients and providers inside atspi-types, is this right? If yes, then atspi-types might not be the best name, maybe atspi-common or atspi-cor would be better?

I'd be happy with atspi-common.

TTWNO:
I'm splitting the crate up further. `atspi-proxies`, `atspi-connection`, `atspi-common`, then those are re-exported via the `atspi` crate via feature flags.

Trying to figure tests out when a doctest requires a certain feature flag has been very painful—this should solve the issue.

TTWNO:
I'm very happy with the organization of these crates like this:

* `atspi-connection` holds the actual connection and communication implementations. Since this is used both in client and future provider implementations, it didn't seem appropriate in either crate on its own.
* `atspi-proxies` holds the (mostly) `zbus-xmlgen` implementations.
* `atspi-common` contains base types for events and communication, without implementing the communication itself.
* `atspi-client` contains the `*_ext.rs` files previously in the primary `atspi` crate behind a feature flag; these implementations add special functionality only useful for clients. Presumably, `atspi-provider` will eventually become a crate as well.

This does make organization a bit more complicated, but has a few big payoffs:

1. Basic types like events and types that can be received by querying AT-SPI can be used without dependence on a heavy crate like `zbus`. (i.e., the types could be included in a restricted environment like WASM).
2. Allows more compilation parallelization.
3. More extensible in the future: simply add an `atspi-*` crate with a feature flag in the primary `atspi` crate to enable it.

Still working on fixing the last bits of semver compliance here, but other than that I'm fairly happy with this branch. If we can start a review process @DataTriny , that would be great! Looking to merge as soon as possible since the attached issues are blocking a handful of features over on Odilia.

DataTriny:
Please rebase and update enumflags2.

TTWNO:
Enumflags2 was downgraded in #87 are you wanting me to reverse that change?

TTWNO:
Also, I have merged main into this branch instead of rebasing, since that was the recommended step from the Github interface. I can also rebase if you'd like that better.

TTWNO:
Ah ha! Finally found the culprit!

I probably need to move a few more things to be workspace deps, but now the tests finally work! Only took a dozen tries :laughing:

TTWNO:
Fixed enumflags2; I didn't realize that the version had been upgraded not downgraded. My bad.

P.S.: This code coverage report is super clear now that we can see the different subcrates: https://app.codecov.io/gh/odilia-app/atspi/pull/82/tree

TTWNO:
I know this breaks the CI, but could we merge this? The error is a sign of another bug that needs to be addressed.

I'll help merge the other PRs once this is merged.

TTWNO:
I'm going to merge this. This PR is blocking too many other PRs, development of Odilia, and general progress. I'm happy to continue submitting smaller PRs after this, but I'm happy with this in its current state, and it has all the changes applied to main since the branch.

Issue #83, opened 2023-05-28
Cache proxy only supports "new" `GetItems` signature


Ablu:
The crate expects `a((so)(so)(so)iiassusau)` as signature. But at least at-spi2-core is still exporting the old version of `a((so)(so)(so)a(so)assusau)` (probably for compatibility): https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/b10fcf21b1ef49dd3d6297ac657434ece3b23578/registryd/registry.c#L864-876

Reference to comment that explains the history: https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/6364851c4b92ab4ed3a3abb0af654f7a5dc9c519/xml/Cache.xml#L72-78

So probably both versions should be supported.

Reproducer:
```rs
use std::error::Error;

use atspi::{cache::CacheProxy, zbus::zvariant::ObjectPath};

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    let atspi = atspi::AccessibilityConnection::open().await?;
    let cache = CacheProxy::builder(atspi.connection())
        .destination(atspi.destination())?
        .path(ObjectPath::from_static_str("/org/a11y/atspi/cache")?)?
        .build()
        .await?;
    let items = cache.get_items().await?;
    dbg!(items);
    Ok(())
}
```

Error:
```
Error: Variant(SignatureMismatch(Signature("a((so)(so)(so)a(so)assusau)"), "`a((so)(so)(so)iiassusau)`"))
```

luukvanderduim:
Thanks for the detailed report.

This is interesting in the sense that at-spi2's registryd returns `CacheItem`s in a layout they deprecated 8 years ago. Maybe for compatibility, but keeping it this way at-spi more or less justifies QT for not moving to the protocols stipulated layout.
But it also goes the other way arround. The fact that Qt still uses it, makes perhaps that AT-SPI2 is more or less forced?
However they should be notified as this interface is not too often used, nevertheless the use it has is really important.

We likely want a fix to support both signatures, I am looking into how to fix this as we speak.

I am also curious to hear about what items `registryd` has cached. 
If these are a collection of items of all servers on the bus, that would be nice for ATs which want to initialize their own cache.

PR #84, opened 2023-06-06
Introduce support for legacy `CacheItems` (issue #83)

luukvanderduim:
Introduce support for legacy `CacheItems`.
Addresses issue #83

Registryd and presumably Qt applications implementing the `Cache` interface return a legacy formatted type when method `GetItems` is called.
    
This adds:
    - A `LegacyCacheItem` type to match the `((so)(so)(so)a(so)assusau)` signature.
    - Add `get_legacy_items` to the `Cache` interface.
    - Add support for `LegacyCacheItem` events (Qt) on the bus.
This is a breaking change, because the enum `CacheEvents` lacks the `#[non_exhaustive]` attribute.
This means that atspi users who match on `CacheEvents` will experience breakage. 
    
The reproducer by @Ablu, when linked against this branch and `get_items` replaced with `get_legacy_items` no longer yields an error:
    
```Bash
Running `target/debug/cache_reproducer`

[src/main.rs:14] items = []
```
    
Receiving Qt-cache signals needs to be tested still, but I would not know of any applications that emit `LegacyCacheItem`s.


TTWNO (approved):

TTWNO:
Thank you for your continued support for this project, @luukvanderduim ! Seriously, I could not do it without you!

PR #85, opened 2023-06-06
Remove non-AT logic.

luukvanderduim:
The XML protocol describes the protocol for both sides on the bus: receivers (ATs) and the senders (accessible applications). `Atspi` generated interfaces based on the XML descriptions. `Atspi` users (ATs) only need the bus- functionality for actions initiated on the AT side, method calls or property inquiries. 
Receiving signals is done by listening to the bus, and matching messages by signature to respective types.

If users do need to send signals on behalf of applications, they  are better catered by [AccessKit]().

This removes 61 signal definitions from the interfaces.

This reduces compile times:
	from ~23.5s to ~21.5s on release builds
	from ~16.5s to ~14s on debug builds,
(Rust nightly 20230605 on a Ryzen 5700G )

This will also shrink the library size:

In release mode:
With DBus  sending generated
-rw-rw-r--   2 luuk luuk 19306826 jun  5 22:10 libatspi.rlib

Without DBus signal sending generated
-rw-rw-r--   2 luuk luuk 13647732 jun  5 22:38 libatspi.rlib

5659094 bytes saved!  \o/

Note 0: Stripping and LTO-ing is left to the end user.
I would not expect the ATs that employ LTO to see  any sizeable difference in this respect.

Note 1: Receiving signals is left untouched by this patch.


DataTriny:
This could be useful as a temporary measure. I'd love to merge AccessKit's AT-SPI provider implementation here at some point though. With #82 introducing a client specific crate, this would be even more logical.

luukvanderduim:
> This could be useful as a temporary measure. 

Maybe. but I dug a little deeper:
#[atspi_proxy] / #[dbus_proxy] (unlike #[dbus_interface])

Makes signal streams out of the signals annotated methods. 
In our current implementation these streams yield nothing, probably unless the user first registers for the given event.
Maybe it is a good idea to automate that in the `#[atspi_proxy(signal)]` attribute? But if we register there on behalf of the user, I worry about deregistering. Making it deregister on drop requires some more gymnastics.

What we currently do to receive signals is fine and this patch would unclutter our API greatly.  But I have to admit my mistake, we did not have 'not-AT logic' around. 

> I'd love to merge AccessKit's AT-SPI provider implementation here at some point though. 

As long as we keep it in seperate subcrates, I think it it would be fine. Not both APIs of services and ATs in one pile, that would be confusing.

> With #82 introducing a client specific crate, this would be even more logical.

Separation of purpose!

TTWNO:
Love this PR! I'm looking to merge #82 soon and would like to see how this can work with it.

In that branch, I've introduced an `atspi-proxies` crate that actually has this generated code. Your proposal is to completely remove this code that sends signals (called events in the `atspi-common` crate); I would suggest that perhaps it could be held behind a feature flag? You'd then have the `atspi-proxies` crate that could have the `signals` and `methods` feature flags to enable different feature sets of the automatically generated proxies.

Perhaps Arnold can clarify if AccessKit uses these signal methods internally to send signals over DBus?

luukvanderduim:
> Love this PR! I'm looking to merge #82 soon and would like to see how this can work with it.
> 
> In that branch, I've introduced an `atspi-proxies` crate that actually has this generated code. Your proposal is to completely remove this code that sends signals (called events in the `atspi-common` crate); I would suggest that perhaps it could be held behind a feature flag? You'd then have the `atspi-proxies` crate that could have the `signals` and `methods` feature flags to enable different feature sets of the automatically generated proxies.
> 
> Perhaps Arnold can clarify if AccessKit uses these signal methods internally to send signals over DBus?

Thanks but it the title of this PR is a bit of a misnomer at this point. I partly misunderstood what was going on. 

I realized later there is also `#[dbus_interface]` aside `#[dbus_pr oxy]` which in our case is `#[atspi_proxy]`.

The [`#[dbus_proxy(signal)]`](https://docs.rs/zbus/latest/zbus/attr.dbus_proxy.html#signals) attribute on a trait method makes the macro create a stream for the methods return type. 
These streams do not work in atspi as-they-are, I am pretty sure you need to register the event beforehand and it is possible we would need to set `default_service="org.a11y.atspi.Registry"` on the trait, otherwise it might be listening to the `DBusProxy` where it will get none. So the streams could work, but we already have a means to receive signals:
 
We have created one stream for all signals and select for signal of interest by registering for those.

So I think it would be sensible to feature gate the signal streams, as you propose, and at least also set `default_service` on all interfaces to make sure the streams come from the Registry. We should also document the the streams that users need to still register / deregister for that particular event.
In `atspi`'s current shape, if users follow the examples, they would not use the individual streams but rather our big stream to rule em all. But if it is feature-gated, they at least could. win-win.

TTWNO:
> So I think it would be sensible to feature gate the signal streams, as you propose, and at least also set default_service on all interfaces to make sure the streams come from the Registry. We should also document the the streams that users need to still register / deregister for that particular event.

Yes. Absolutely.

> In atspi's current shape, if users follow the examples, they would not use the individual streams but rather our big stream to rule em all. But if it is feature-gated, they at least could. win-win.

Yes.

> The [#[dbus_proxy(signal)]](https://docs.rs/zbus/latest/zbus/attr.dbus_proxy.html#signals) attribute on a trait method makes the macro create a stream for the methods return type.

Ahhh that explains the huge advantage in compile time.

DataTriny:
> Perhaps Arnold can clarify if AccessKit uses these signal methods internally to send signals over DBus?

No we don't. It is way more efficient to send signals directly using the lower-level zbus API, especially since these signals don't belong to regular AT-SPI interfaces, but rather to `org.a11y.atspi.Event.*` ones.

TTWNO:
> No we don't. It is way more efficient to send signals directly using the lower-level zbus API, especially since these signals don't belong to regular AT-SPI interfaces, but rather to org.a11y.atspi.Event.* ones.

Hmmmm, perhaps we can just remove this entirely then...

TTWNO:
Confirmed your results on my machine.

TTWNO:
This only decreases code coverage in atspi_macros because it doesn't hit those "do X in case of signal" cases. Don't worry about that.

TTWNO:
This is a pretty simple change, with pretty big implications for binary size. I'm very happy with it. Thanks, @luukvanderduim !

luukvanderduim:
Yes this was text-book low-hanging fruit.

luukvanderduim:
Thanks for merging!

PR #79, opened 2023-05-04
Update XML from upstream

TTWNO:

PR #86, opened 2023-06-08
Update xml and code accordingly

luukvanderduim:
There have been a few changes in upstream XML protocol descriptions since last time we updated.

Changes are found in: Accessible.xml, Component.xml and TableCell.xml.

Mostly typos, but also a few changes to the API.
TableCell gains the methods GetColumnHeaderCells and GetRowHeaderCells.

Three commits:
- Add changes to XML files
- Add methods to TableCell interface
- A handful of Clippy / spelling fixes.

### DIFF
--- a/src/accessibility_connection.rs
+++ b/src/accessibility_connection.rs
@@ -251,7 +251,7 @@ impl AccessibilityConnection {
 	}
 
 	/// Shorthand for a reference to the underlying [`zbus::Connection`]
-	#[must_use]
+	#[must_use = "The reference to the underlying zbus::Connection must be used"]
 	pub fn connection(&self) -> &zbus::Connection {
 		self.registry.connection()
 	}
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -26,7 +26,7 @@ pub struct CacheItem {
 	pub index: i32,
 	/// Child count of the accessible  i
 	pub children: i32,
-	/// The exposed interfece(s) set.  as
+	/// The exposed interface(s) set.  as
 	pub ifaces: InterfaceSet,
 	/// The short localized name.  s
 	pub short_name: String,
--- a/src/error.rs
+++ b/src/error.rs
@@ -61,7 +61,7 @@ impl std::fmt::Display for AtspiError {
 		match self {
 			Self::Conversion(e) => f.write_str(&format!("atspi: conversion failure: {e}")),
 			Self::MemberMatch(e) => {
-				f.write_str(format!("atspi: member mismatch in conversion: {e}").as_str())
+				write!(f, "atspi: member mismatch in conversion: {e}")
 			}
 			Self::UnknownBusSignature(s) => {
 				write!(f, "atspi: Unknown bus body signature: {s}")
--- a/src/error.rs
+++ b/src/error.rs
@@ -63,7 +63,9 @@ impl std::fmt::Display for AtspiError {
 			Self::MemberMatch(e) => {
 				f.write_str(format!("atspi: member mismatch in conversion: {e}").as_str())
 			}
-			Self::UnknownBusSignature(s) => f.write_str("atspi: Unknown bus body signature: {s}"),
+			Self::UnknownBusSignature(s) => {
+				write!(f, "atspi: Unknown bus body signature: {s}")
+			}
 			Self::UnknownInterface => f.write_str("Unknown interface."),
 			Self::MissingInterface => f.write_str("Missing> interface."),
 			Self::UnknownSignal => f.write_str("atspi: Unknown signal"),
--- a/src/error.rs
+++ b/src/error.rs
@@ -12,8 +12,9 @@ pub enum AtspiError {
 	/// On specific types, if the event / message member does not match the Event's name.
 	MemberMatch(String),
 
-	/// To indicate a match or equality test on a signa body signature failed.
-	UnknownBusSignature,
+	/// To indicate a match or equality test on a signal body signature failed.
+	/// The contained string is the signature that was found.
+	UnknownBusSignature(String),
 
 	/// When matching on an unknown interface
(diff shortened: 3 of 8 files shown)


DataTriny (approved):

DataTriny:
This completely supersedes #79 right?

DataTriny:
Let's keep `assume_defaults` out of this discussion, it's not really linked to the update of AT-SPI schema anyway.

I think most AT-SPI providers don't set the destination for the signals they emit. If you set it to the registry, then I'm almost certain that only the registry will receive them. That's not what we want. In the peer-to-peer setup (that we don't yet support), the registry is not even registered on the private bus.

In most cases you indeed want to send your messages directly to the provider, not the registry.

And, as I said above, some providers even support peer-to-peer communication, to eliminate the bus daemon from the equation.

luukvanderduim:
> Let's keep `assume_defaults` out of this discussion, it's not really linked to the update of AT-SPI schema anyway.

Its not important.

> I think most AT-SPI providers don't set the destination for the signals they emit.

This means they are broadcasted.

> If you set it to the registry, then I'm almost certain that only the registry will receive them. That's not what we want.

That is not the impression I get.
ATs have to register for events with the Registry. We have to do both, set MatchRules and register for an events with registry.
 `MatchRule`'s are used to inform providers that there is interest in a given signal, to reduce work on their part.
I (and @TTWNO ) tried setting a MatchRule only and listen for events with the DBusProxy, but that was unsuccessful.
Sp I suspect providers send to Registry only and that is why we need to register, but to be really sure I would need to read and understand the registryd code.

Anyway that all is off-topic too.

> In the peer-to-peer setup (that we don't yet support), the registry is not even registered on the private bus.

That is an appealing prospect, but P2P is for another day. Lets first improve on what we have. 

> And, as I said above, some providers even support peer-to-peer communication, to eliminate the bus daemon from the equation.

On the a11y bus? Which applications / toolkits you know of offer P2P?

DataTriny:
> Which applications / toolkits you know of offer P2P?

At least GTK3/ATK apps implemented this.

> On the a11y bus?

No. These are privase buses spawned by each application. You have to get the bus address by calling `GetApplicationBusAddress` on the `org.a11y.atspi.Application` interface.

DataTriny:
Does anyone have an idea why an event unit test failed on the coverage CI job?

luukvanderduim:
No. Signatures are fine and tests run fine locally. Cosmic rays in the test environment?

luukvanderduim:
```Rust
	Ok(e) => {
		println!("{e:?}");
		panic!("Something else happened");
	}
```
The `e` here is an `Option<Result<Event, AtspiError>>`, the println does not show up in the test log because printlns are always suppressed in tests.
The problem is with the Cache interface and the test runs fine locally.. Odd.

luukvanderduim:
This is interesting.. `TryFrom<Message> for Event` returns `UnknownBusSignature`, the bus signature that is not recognized is "s". AFAIK there is no (signal) signature on the bus with signature "s". We expected "(so)" here.

luukvanderduim:
Registering and setting  `MatchRule` before constructing the stream likely fixed this problem.

DataTriny:
Could this be `NameAcquired`/`NameLost` signals?

Codecov is a bit annoying here...

DataTriny (approved):

luukvanderduim:
> Could this be `NameAcquired`/`NameLost` signals?

Very good catch!

We had all signals in all interface XMLs covered. But for the DBus interface, we rely on `zbus`' own implementation, therefore we had no such XML in the `xml/` directory and these two ( plus `NameOwnerChanged` (sss) ) escaped.

> Codecov is a bit annoying here...

I added two methods which increases untested surface.

euclio:
Hi folks, sorry to bump this PR, but I'm running into the same `UnknownBusSignature` issue on my local machine using 0.17. It's coming from a `NameAcquired` event. It's not clear to me what the issue is: do I need to pull in newer atspi code?

luukvanderduim:
Hi Andy,

Thanks for the bump. 
I had not yet come around to adding these signals. I believe there are only two of these I believe.

For the time being you could safely ignore them and or obtain those signals using the `zbus::fdo::DBusProxy`. Is your plan to monitor for those signals through atspi's `signal_stream`? 

I will hurry up implementing them.

euclio:
Honestly, I don't really know what I'm doing... :)

Thanks for the info though. I'm completely new to AT-SPI, I'm experimenting with using this crate to drive some automated UI tests for my GTK application.

luukvanderduim:
There are four `DBusProxy` signals which not cover. These come from the DBus daemon, in most cases nowadays that is `dbus-broker`. 
These are the signals:
`NameLost`, `NameAcquired`, `NameOwnerChanged` and `PropertiesChanged`

You can look them up in the [DBus specs](https://dbus.freedesktop.org/doc/dbus-specification.html) to see what they mean and if you wish to monitor those signals. I guess it depends on the AT and what it requires to know about the bus servers. 

I could add them, however - on the one hand, its nice to handle all possible signals - on the other hand, these signals have little to do with AT-SPI itself and these are currently covered through zbus' `DBusProxy`. 

I am curious to know if you had been wanting to monitor these in the `event_stream`? If so, let's make a new issue to add them. (This one is actually a bit unrelated)

luukvanderduim:
> Honestly, I don't really know what I'm doing... :)
> 
> Thanks for the info though. I'm completely new to AT-SPI, I'm experimenting with using this crate to drive some automated UI tests for my GTK application.

Hop in our [matrix / discord chat](https://github.com/odilia-app/odilia#community)  if you have any questions. I'll be glad to help out if I have an answer that is.

TTWNO:
This update has not been published to `crates.io`, which is probably why you may still be experiencing problems.

PR #87, opened 2023-06-10
Update Cargo.toml

Dylan-DPC:
Updates dependency to the patch version of 0.7.7

[Advisory](https://rustsec.org/advisories/RUSTSEC-2023-0035.html)
[fix](https://github.com/meithecatte/enumflags2/releases/tag/v0.7.7)

(Side note, you don't need to explicitly mention ^ in your dependencies, since cargo does `^` implicitly if just the value is mentioned. I've kept it on for consistency sake but it can be removed form all if needed)


DataTriny (approved):

DataTriny:
Thank you @Dylan-DPC

PR #88, opened 2023-06-15
Fix async test assumptions

luukvanderduim:
The `Cache` recv_add and recv_remove tests were flaky in CI because the test assumed that after sending the event, that event should be the first it would receive, which was not always true.

This fixes that by using a timeout stream (instead of a timeout on a single future).
These tests now handle all events within a span of a second to receive the event of interest.
This works fine.

The other thing this PR adds are some fixes with respect to outer parentheses on DBus signatures:

[DBus omits outer parentheses on the wire.](https://docs.rs/zbus/latest/zbus/struct.Message.html#method.body_signature)
This means that if we read a DBus signature "ii", "(ii)" is implied.

This adds:
    - `assert_eq_signatures` macro that checks equality regardless of outer parentheses.

This PR also fixes signatures where we compare with `Message::body_signature()`, that is in our stream parsing `TryFrom<Message> for Event`

When the `Type::signature`'s on the user facing types are checked, it will show outer parentheses again.

Note that others on the bus deserialize stripped signatures just fine and that those `Type::signatures` do include outer parentheses.

With this merged, that problem is fixed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "atspi"
 version = "0.15.1"
-authors = ["Michael Connor Buchan <mikey@blindcomputing.org>", "Tait Hoyem <tait@tait.tech>", "Alberto Tirla <albertotirla@gmail.com>", "DataTriny <datatriny@gmail.com>", "Luuk Duim <luukvanderduim@gmail.com>"]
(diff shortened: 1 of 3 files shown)


TTWNO:
Is this branch still necessary? I thought it was solved by another one of your PRs?

luukvanderduim:
You are right, it has no use anymore. This was the same thing, against pre-refactor main.

PR #91, opened 2023-06-17
Error serialization

TTWNO:
Allow the `AtspiError` type to be serialized with serde.

This "removes" some error types that used to be nested within the `AtspiError` type, but realistically we weren't really handling those errors specifically anyway, so it shouldn't be a big deal.

If it comes up that a nested event within `zbus::Error`, `zbus::fdo::Error`, `zvariant::Error`, or `zbus_names::Error` arises, we can always convert it explicitly in the `From<T> for AtspiError` implementation blocks.


TTWNO:
Since this mostly changes the type of `AtspiError`, there really isn't much to test, even if technically it adds a bunch of untested lines.

atspi-common/README.md
```diff
@@ -5,3 +5,19 @@ This crate is meant to only contain the absolute lowest common denominator for l
 therefore, it attempts to be able to compile on basically any architecture (including WASM).
 
 Please use the internal documentation to learn how to use these data structures.
+
+## Compiling for WASM
```
DataTriny, line 9: How error serialization and Wasm support are related? Please make the job of reviewers easier by keeping your PRs focused on one new feature.

Thanks.

DataTriny (changes requested):

atspi-common/README.md
TTWNO: To send Rust types across WASM FFI boundaries, tou need to e bl to serialize to some stable gormat, then deserialize on the other side.

TTWNO:
Although this allows `AtspiError` to be serialized... it gives up too much in terms of error opaqueness.

PR #92, opened 2023-06-18
Fix #83 again (now with tests)

### LINKED ISSUE #83: Cache proxy only supports "new" `GetItems` signature

TTWNO:
…rsion functions, and tests

### DIFF
--- a/atspi-common/src/events/mod.rs
+++ b/atspi-common/src/events/mod.rs
@@ -137,7 +137,12 @@ pub struct LegacyAddAccessibleEvent {
 	pub item: Accessible,
 	pub node_added: LegacyCacheItem,
 }
-impl_event_conversions!(LegacyAddAccessibleEvent, CacheEvents, CacheEvents::LegacyAdd, Event::Cache);
+impl_event_conversions!(
+	LegacyAddAccessibleEvent,
+	CacheEvents,
+	CacheEvents::LegacyAdd,
+	Event::Cache
+);
 event_test_cases!(LegacyAddAccessibleEvent);
 impl_from_dbus_message!(LegacyAddAccessibleEvent);
 impl_to_dbus_message!(LegacyAddAccessibleEvent);
--- a/atspi-common/src/events/mod.rs
+++ b/atspi-common/src/events/mod.rs
@@ -32,7 +32,7 @@ use zbus_names::{OwnedUniqueName, UniqueName};
 use zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Signature, Type, Value};
 
 use crate::{
-	cache::CacheItem,
+	cache::{CacheItem, LegacyCacheItem},
 	error::ObjectPathConversionError,
 	events::{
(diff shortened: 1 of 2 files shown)


atspi-common/src/events/mod.rs
```diff
@@ -542,6 +584,11 @@ impl TryFrom<&zbus::Message> for Event {
 				let ev = AddAccessibleEvent::try_from(msg)?;
 				Ok(Event::Cache(CacheEvents::Add(ev)))
 			}
+			// LegacyCacheAdd signature
+			"((so)(so)(so)a(so)assusau)" => {
```
luukvanderduim, line 588: If you look at the top of this `TryFrom<Message> for Event` impl, you will see we are comparing to a 'Marshalled' signature, because that is what `body_signature()` returns. It has no outer parentheses, typically one pair less than our unmarshalled / Rust type ``Type::Signature()` ones.
In other words, I think we do not want the outer parentheses here, also goes for the other signatures here.

atspi-common/src/events/mod.rs
```diff
@@ -126,9 +126,51 @@ pub enum Event {
...
+	}
+
+	fn sender(&self) -> UniqueName<'_> {
+		self.item.name.clone().into()
+	}
+	fn path(&self) -> ObjectPath<'_> {
+		self.item.path.clone().into()
+	}
+	fn body(&self) -> Self::Body {
+		(self.node_added.clone(),)
```
luukvanderduim, line 170: Those outer parentheses are unneeded. No need to compensate for the D-Bus marshalling.

luukvanderduim (changes requested):
Please remove the unneeded parentheses and we are done!

luukvanderduim (approved):

PR #93, opened 2023-06-18
A new port of cache events

TTWNO:


atspi-common/src/events/mod.rs
```diff
@@ -42,6 +42,30 @@ use crate::{
 };
 //use atspi_macros::try_from_zbus_message;
 
+pub(crate) fn signatures_are_eq(lhs: &Signature, rhs: &Signature) -> bool {
```
TTWNO, line 45: Please add some simple tests for this function. It *is* already covered by the existing, more complex tests, but if an issue is found, I'd rather a more specific test fails.

atspi-common/src/events/mod.rs
```diff
@@ -71,6 +95,18 @@ pub struct EventBodyQT {
 	pub properties: (String, OwnedObjectPath),
 }
 
+impl Default for EventBodyQT {
```
TTWNO, line 98: Why have you manually implemented `Default`?

atspi-common/src/events/mod.rs
```diff
@@ -99,6 +135,18 @@ impl From<EventBodyQT> for EventBodyOwned {
 	}
 }
 
+impl Default for EventBodyOwned {
```
TTWNO, line 138: Similar question: Why have you implemented `Default` manually here?

atspi-common/src/macros.rs
```diff
@@ -256,14 +256,14 @@ macro_rules! event_wrapper_test_cases {
...
-					"/org/a11y/sixynine/fourtwenty",
+					"/org/a11y/sixtynine/fourtwenty",
 					"org.a11y.atspi.technically.valid.lol",
 					<$any_subtype as GenericEvent>::DBUS_MEMBER,
 				)
 				.unwrap()
 				.sender(":0.0")
 				.unwrap()
-				.build(&(<$any_subtype>::default().body()))
+				.build(&<$any_subtype>::default().body())
```
TTWNO, line 266: This is a great change! Not sure how I missed this.

atspi-common/src/macros.rs
```diff
@@ -353,3 +353,41 @@ macro_rules! event_test_cases {
 		assert_impl_all!(zbus::Message: TryFrom<$type>);
 	};
 }
+
+/// Asserts that the signatures are equal, but ignores the outer parentheses as
+/// the difference between marshalled and unmarshalled signatures is often just one set of outer parentheses.
+#[macro_export]
+macro_rules! assert_eq_signatures {
```
TTWNO, line 360: I see why this is implemented separately from the plain function, but it's unfortunate that it has a completely separate code path. Any ideas here?

atspi-common/src/events/mod.rs
luukvanderduim: Cannot derive if not all fields implement `Default` and [`Value`](https://docs.rs/zvariant/3.14.0/zvariant/enum.Value.html) hasn't.

atspi-common/src/macros.rs
luukvanderduim: The excuse is because I wrote the macro before I wrote the function and it never occurred to me to implement the macro in terms of the function. Its a good idea! Thanks!

luukvanderduim: [x]   implement `assert_eq_signatues!` in terms of `signatures_are_eq()`.

atspi-common/src/macros.rs
```diff
@@ -358,36 +358,7 @@ macro_rules! event_test_cases {
...
-		match (bus_sig_has_outer_parens, type_sig_has_outer_parens) {
-			(true, false) => {
-				assert_eq!(lhs_sig.slice(1..lhs_sig.len() - 1).as_bytes(), rhs_sig.as_bytes());
-			}
-			(false, true) => {
-				assert_eq!(lhs_sig.as_bytes(), rhs_sig.slice(1..rhs_sig.len() - 1).as_bytes());
-			}
-			_ => assert_eq!(lhs_sig.as_bytes(), rhs_sig.as_bytes()),
-		}
+	($lhs_sig:expr, $rhs_sig:expr) => {
```
TTWNO, line 361: The error message should include the values being compared, since right now this would simply show `thread panicked: assert!(false)`.

luukvanderduim (approved):

PR #98, opened 2023-06-24
Use a state type for the state field in StateChangedEvent

TTWNO:
Be more strict about the `state` field.

### DIFF
--- a/atspi-common/README.md
+++ b/atspi-common/README.md
@@ -5,3 +5,19 @@ This crate is meant to only contain the absolute lowest common denominator for l
 therefore, it attempts to be able to compile on basically any architecture (including WASM).
 
 Please use the internal documentation to learn how to use these data structures.
+
+## Compiling for WASM
+
+If you'd like to compile this crate for any architecture that implements `std`, then you have to compile it without any feature flags.
+By default, this crate provides a variety of converison methods for foreign error types into its own event type.
+
+## Feature Flags
+
+* `async-std`: build `zbus` with `async-std` support.
+* `tokio`: build `zbus` with `tokio` support.
+
+Either of these flags will enable conversion from zbus error types to the `AtspiError` type.
+
+> Why do I need to specify a runtime to have the conversion methods available to me?
+
+A: Because `zbus` requires you to specify at least one runtime, and it would be annoying to compile an entire second runtime if it is not necessary.
--- a/atspi-common/src/error.rs
+++ b/atspi-common/src/error.rs
@@ -1,9 +1,12 @@
 #[allow(clippy::module_name_repetitions)]
 #[derive(Debug, serde::Serialize, serde::Deserialize, Eq, PartialEq, Clone)]
+#[non_exhaustive]
 /// An error type that can describe atspi and `std` and different `zbus` errors.
 pub enum AtspiError {
-	/// When testing on either variant, we might find the we are not interested in.
-	CacheVariantMismatch,
+	/// An error for when you attempt to extract the incorrect variant of a type.
+	/// For example: [`crate::events::Event`] has many variants and sub-variants that can be extracted.
+	/// If you attempt to convert to a variant which is not the variant contained within the `Event` enum, then you will get this error.
+	TypeMismatch,
 
 	/// On specific types, if the event / message member does not match the Event's name.
 	MemberMatch(String),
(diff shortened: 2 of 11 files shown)


TTWNO:
I had implemented Default for State, setting Invalid as the default option of none others are available. Unfortunately, this isn't really correct since "invalid" getting sent over the bus, and a state of invalid characters (for example "fire"), should not have the same behaviour.

TTWNO:
Blocked on #91

atspi-common/README.md
```diff
@@ -5,3 +5,19 @@ This crate is meant to only contain the absolute lowest common denominator for l
 therefore, it attempts to be able to compile on basically any architecture (including WASM).
 
 Please use the internal documentation to learn how to use these data structures.
+
+## Compiling for WASM
+
+If you'd like to compile this crate for any architecture that implements `std`, then you have to compile it without any feature flags.
+By default, this crate provides a variety of converison methods for foreign error types into its own event type.
```
luukvanderduim, line 12: conversions

atspi-common/src/state.rs
```diff
@@ -219,6 +221,122 @@ pub enum State {
 	ReadOnly,
 }
 
+impl From<State> for String {
```
luukvanderduim, line 224: note you added this conversion to be able to print it elsewhere, then maybe save a conversion and `impl Display`, but you perhaps have other uses in mind as well..

atspi-common/src/state.rs
```diff
@@ -11,10 +12,11 @@ use zvariant::{Signature, Type};
 /// of an accessibility object.
 #[bitflags]
 #[repr(u64)]
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Default)]
```
luukvanderduim, line 15: Love `Default`. Hash is here to help users storing states in a `HashMap`?

luukvanderduim (commented):
Much good stuff.
Though I am unsure about making the error more opaque. I cannot see why that was necessary while on the other hand I could see why a user wants to sift types of error by origin?

atspi-common/src/error.rs
```diff
@@ -30,69 +26,72 @@ pub enum AtspiError {
...
-	UnknownSignal,
+	UnknownSignal(String),
 
 	/// Other errors.
 	Owned(String),
 
 	/// A `zbus` or `zbus::Fdo` error. variant.
 	Zbus(String),
 
-	/// A `zbus_names` error variant
```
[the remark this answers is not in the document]
TTWNO: `zvariant` and `zbus_names` error types can not be serialized. My primary issue is that I need to be able to send errors from/to a WASM guest. Eventually, generic reference types should be able to be sent over a WASM FFI boundary, but that is still some months/years out.

I could make a PR to those two repos for a `serde` flag, that would enable serialization of these types. What do you think about that?

atspi-common/src/events/mod.rs
```diff
@@ -555,7 +549,7 @@ impl TryFrom<Event> for AvailableEvent {
 		if let Event::Available(specific_event) = generic_event {
 			Ok(specific_event)
 		} else {
-			Err(AtspiError::Conversion("Invalid type"))
+			Err(AtspiError::InvalidType)
```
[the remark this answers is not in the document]
TTWNO: `TypeMismatch` coming up!

atspi-common/src/state.rs
TTWNO: Technically it is still for conversion. Basically, I just want to infallibly convert to a String for a conversion from a `Message` to a `T: GenericEvent`. I could unwrap a `serde` call here, but just not sure how I feel about that.

atspi-common/src/state.rs
TTWNO: Using a State as a *key* in a HashMap.

atspi-common/src/error.rs
```diff
@@ -1,11 +1,7 @@
 #[allow(clippy::module_name_repetitions)]
-#[derive(Debug)]
-#[non_exhaustive]
+#[derive(Debug, serde::Serialize, serde::Deserialize, Eq, PartialEq, Clone)]
 /// An error type that can describe atspi and `std` and different `zbus` errors.
 pub enum AtspiError {
-	/// Converting one type into another failure
-	Conversion(&'static str),
-
 	/// When testing on either variant, we might find the we are not interested in.
```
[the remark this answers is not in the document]
TTWNO: Not sure what you're commenting on. This was a removed error variant that could not be serialized.

atspi-common/src/error.rs
```diff
@@ -1,11 +1,7 @@
 #[allow(clippy::module_name_repetitions)]
-#[derive(Debug)]
-#[non_exhaustive]
```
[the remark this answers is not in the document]
TTWNO: Not sure how `#[non_exhaustive]` got removed. That's on me. Put it back in the latest commit.

atspi-common/src/state.rs
luukvanderduim: Maybe don't do that. There are still `Message`s which we do not parse to an event. The two DBus messages with "s" signature. (forgot their names)
And if interfaces change, or there are signals we somehow don't yet know of, we do not want to panic unless we absolutely must. We want ATs to handle an error and carry on.

PR #99, opened 2023-06-24
WASM CI Check

TTWNO:
Add a CI check for the common crate, mandating that it at least compiles for WASM. We can not run the tests for a WASM target because many tests require the use of `zbus`, which can not compile to WASM, yet.

This PR is also blocked on #91 .


TTWNO:
This will need to be re-created not based on #91

PR #100, opened 2023-07-02
Add more tests, remove old comments

TTWNO:


atspi-common/src/interface.rs
```diff
@@ -52,26 +73,26 @@ impl<'de> Deserialize<'de> for Interface {
 				E: de::Error,
 			{
 				match value {
```
luukvanderduim, line 75: Why replace the literals "org.a11y.atspi.Event.Accessible", with constants.

What is the advantage over implementing `Display` on the `Interface` enum and use that variant where we need it?
 
I know it was not in this commit but would you know why a manual implementation of `Serialize` and `Deserialize`? Trying to think of a reason but if we had `TryFrom<&str> for Interface` and the `From<Interface> for &str` would that allow for deriving? 

Can we try? If possible I would choose derived serde impls over manual ones.

luukvanderduim (commented):
One question on the constants and the `Interface` enum's manual imps. 
The latter half (manual impls) is out of scope for this PR so we _can_ defer the question but it would be nice if we thought about it for a bit.

atspi-common/src/interface.rs
TTWNO: > Trying to think of a reason but if we had TryFrom<&str> for Interface and the From<Interface> for &str would that allow for deriving?

It would still require a manual implementation of `serde`'s Serialize and Deserialize traits, it's just that we'd be able to take advantage of existing conversions.

> I know it was not in this commit but would you know why a manual implementation of Serialize and Deserialize? 

I suspect that it was because there's no easy way to set the value of an enum variant to an arbitrary string with serde?

TTWNO: > Why replace the literals "org.a11y.atspi.Event.Accessible", with constants.

I have had issues before with a mis-matched name in one place not matching another. Using constants avoids any ambiguity. They either work everywhere, or work nowhere.

PR #104, opened 2023-07-03
WASM CI Check

TTWNO:
The common crate should be able to be compiled for WASM for easy integration across many different applications. The `zbus` and `proxies` flags are available for the common crate, just not for the WASM targets.

This will need to be updated as `-preview2` becomes available.


atspi-common/src/state.rs
```diff
@@ -485,4 +595,23 @@ mod tests {
...
+	fn convert_state_direct_string() {
+		for state in StateSet::from_bits(0b11111111111111111111111111111111111111111111).unwrap().iter() {
+			let state_str: String = state.into();
+			let state_two: State = state_str.clone().into();
+			assert_eq!(state, state_two, "The {state:?} was serialized as {state_str}, which deserializes to {state_two:?}");
+		}
+	}
+	#[test]
+	fn convert_state_direct_string_is_equal_to_serde_output() {
+		for state in StateSet::from_bits(0b11111111111111111111111111111111111111111111).unwrap().iter() {
```
luukvanderduim, line 609: Very convenient that the `StateSet` allows us to iterate over it like this.

luukvanderduim (approved):

PR #105, opened 2023-07-05
Session status

luukvanderduim:
This PR features commits which:

- Add `read_session_accessibility`
- Add `AtspiResult<T>`
- Fixes spelling and formats.

We already have a means to set session accessibility status 
with: `set_session_accessibility`, but no way to read its status.

This adds `read_session_accessibility`

Applications can determine whether to activate their at-spi2 features by
reading the `IsEnabled` property on the `Status` proxy.

Reading `false` when an AT is running should be considered a bug.

AT's should set the `IsEnabled` property before anything else.

Reference: [Freedesktop AT-SPI2 FAQ](https://www.freedesktop.org/wiki/Accessibility/AT-SPI2/)

### DIFF
--- a/atspi-common/src/events/mod.rs
+++ b/atspi-common/src/events/mod.rs
@@ -612,7 +612,9 @@ impl TryFrom<&zbus::Message> for Event {
 			},
 			// Atspi / Qspi signature
 			"siiva{sv}" | "siiv(so)" => {
-				let Some(interface) = msg.interface() else {  return Err(AtspiError::MissingInterface);  };
+				let Some(interface) = msg.interface() else {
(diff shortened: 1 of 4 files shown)


TTWNO:
Since we need to create a connection to get or set this, should these just be methods of `AccessibilityConnection`?

luukvanderduim:
> Since we need to create a connection to get or set this, should these just be methods of `AccessibilityConnection`?

Not in favor. These operate on the `Status` interface implemented by the at-spi-bus-launcher daemon, which lives on the session bus.

`AccessibilityConnection` is our handle to the accessibility bus.

PR #106, opened 2023-07-05
 [atspi-common] Docs for RelationType

luukvanderduim:
This documents `RelationType`


luukvanderduim:
> Do we need the `= 0` for the `Null` variant?
>
I don't think we do.
 
> Otherwise these are good docs.

:+1:

PR #108, opened 2023-07-06
Remove `ObjectPair` and `(String, OwnedObejctPath)`, replace with `Accessible`

### LINKED ISSUE #107: Convert `ObjectPair` and `(String, OwnedObject)` into `Accessible`s.
As a library, we want uniform and unambiguous typing to avoid confusion and errors for users and maintainers.

In `atspi` we find `ObjectPair`s, `Accessible`s, and `(String, zvariant::OwnedObject)` tuples which are different representations of the same data.

`ObjectPair` is an alias for `(String, zvariant::OwnedObject)`.

`Accessible` has this form:

```rust
#[derive(Debug, Clone, Serialize, Deserialize, Type, PartialEq, Eq, Hash)]
pub struct Accessible {
	pub name: OwnedUniqueName,
	pub path: OwnedObjectPath,
}
```

In the context of `atspi` we want these to all become `Accessible`'s as that describes their purpose within this domain best.

TTWNO:
Fixes #107 

* [X] Set type of `Accessible.name` to String, since that is what is sent across the bus.
* [X] Move `Accessible` into its own module within `atspi-common`
* [X] Remove all references to `Objectpair`
* [X] Remove all references to `(String, zvariant::OwnedObjectPath)` 

```shell
$ grep -rne '(String, zvariant::OwnedObjectPath)'
$ 
```

```shell
$ grep -rne 'ObjectPair'
$ 
```

### DIFF
--- a/atspi-common/src/accessible.rs
+++ b/atspi-common/src/accessible.rs
@@ -1,7 +1,6 @@
-use crate::{AtspiError, events::signatures_are_eq};
+use crate::events::signatures_are_eq;
 use serde::{Serialize, Deserialize};
 use zvariant::{Type, OwnedObjectPath, ObjectPath, Signature};
-use zbus_names::{OwnedBusName, BusName};
 
 pub const ACCESSIBLE_PAIR_SIGNATURE: Signature<'_> = Signature::from_static_str_unchecked("(so)");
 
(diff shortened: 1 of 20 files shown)


luukvanderduim:
I found one more  `(Stirng, OwnedObjectPath)` extra!

Even though the tests are green, I would appreciate it if we tested this branch in Odilia first before giving it the final thumbs up - or have you done so already @TTWNO ?

TTWNO:
I'm trying to not consider Odilia when it comes to changes to the `atspi` library. Whatever is best for end users of the library is likely also the right choice for Odilia.

PR #109, opened 2023-07-06
[atspi-connection] Use a channel to communicate.

TTWNO:
* Some issues faced by @luukvanderduim when opening Thunderbird caused event storms to overwhelm the bus.
* Since the MessageStream was not being polled fast enough, `zbus` was getting clogged up, which can cause issues like hanging.
	* This is even mentioned in the documentation for `zbus`.
* This commit takes a different approach, adding a channel, where the receiving end is send out to the user (luckily for us, the return type hasn't changed at all thanks to `impl Stream<_>`)
* Now, a separate task will poll the MessageStream, and add it to the channel.
* If the channel gets backed up, we having tracing/debug information avaiable as to why this happened.
* Now, at least if `zbus` is getting backed up, we should be able to find out what is slowing down to polling of the channel, and then blocking up `zbus` indirectly.
* It is possible that we want to spawn a second *thread* to deal with these events storms, but I suspect this will not be necessary unless the "30,000" events reported by Luuk simply overwhelm even `tokio`'s executor.
* Odilia *should* in theory be unaffected by this, since each even handler is spawned on its own task, and Odilia shouldn't back up unless a very large quanitiy of events are mutating the same items in the cache (worst case scenario with locking).
	* The other option is that we simply can't launch tasks fast enough. Based on [these benchmarks](http://gravitext.com/2020/01/13/blocking-permit.html) it seems spawning may be in the hundreds of microseconds, which may not be able to keep up with 30,000 spawns a second (it would need to be at least less than 333us, but ideally we'd want to some headroom).

If possible, I'd like [Luuk's statspi](https://github.com/luukvanderduim/statspi) to be tested with this branch and see if it can stave off the event storms backing up `zbus`s very small queue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -192,11 +192,14 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
-      - name: Install WASM target
-        run: rustup target add wasm32-wasi
+      - name: Install WASM targets
+        run: rustup target add wasm32-wasi && rustup target add wasm32-unknown-unknwon
       - name: Test Common Compilation (wasm32-wasi)
         working-directory: atspi-common
-        run: cargo build --target wasm32-wasi
+        run: cargo build --target wasm32-wasi --no-default-features
+      - name: Test Common Compilation (wasm32-unknown-unknown)
+        working-directory: atspi-common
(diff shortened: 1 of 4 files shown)


TTWNO:
Luuk mentioned that an additional level of indirection would make this much harder to debug. I think he's right.

PR #110, opened 2023-07-07
Enforce some documentation

TTWNO:
* [ ] Enable `#[deny(missing_docs)]` on `atspi-common`
* [ ] Fully document `atspi-common`

EDIT: Currently undocumented items: 77


luukvanderduim:
I am currently documenting `trait Accessible`'s methods and use `Accessible.xml` mainly as source and start from there.
I noticed it has [pretty nice descriptions for `Role` variants](https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/main/xml/Accessible.xml#L772-797) , these descriptions are listed beneath 
`GetRole`

They are really nice, but written from the perspective of interface implementers, so I do rewrite them to reflect our perspective. 

For example if the docs say:
"You should return 4, when A and B" 
I write 
"Expect to receive 4, if conditions A and B are met."

Most time is spent on making sure internal links are correct and changing foreign `ATSPI_ROLE_SOME_ROLE` to [`Role::SomeRole`].

TTWNO:
> They are really nice, but written from the perspective of interface implementers, so I do rewrite them to reflect our perspective.

Ideally no. Since these types in `atspi-common` are meant to be used anywhere, it doesn't make sense to write comments as if we are a server or client. We should, in theory, only have neutral comments.

TTWNO:
> Most time is spent on making sure internal links are correct and changing foreign ATSPI_ROLE_SOME_ROLE to [Role::SomeRole].

It is a huge pain. I had to do it for `Role` variants in `role.rs` and boy gee golly that took almost 3 hours. That said, I bet you €20 that there's a way to define a macro for this.

luukvanderduim:
> > Most time is spent on making sure internal links are correct and changing foreign ATSPI_ROLE_SOME_ROLE to [Role::SomeRole].
> 
> It is a huge pain. I had to do it for `Role` variants in `role.rs` and boy gee golly that took almost 3 hours. That said, I bet you €20 that there's a way to define a macro for this.

This is an evening activity, you can do this stuff without much thinking. Its kindof zen.

luukvanderduim:
> > They are really nice, but written from the perspective of interface implementers, so I do rewrite them to reflect our perspective.
> 
> Ideally no. Since these types in `atspi-common` are meant to be used anywhere, it doesn't make sense to write comments as if we are a server or client. We should, in theory, only have neutral comments.

If the phrasing requires a perspective it should align with our users', not those of AccessKit, QtSpi, Atk.
AccessKit and others who have atspi as dependency surely expect this perspective.

We could, however, where we reasonably can, try a more neutral 'bus-daemon' perspective.

PR #111, opened 2023-07-07
Simplify event marshaling

TTWNO:


TTWNO:
This should probably also re-define some error types so that `UnknwonBusInterface` and `UnknownMethod` can be returned.

PR #115, opened 2023-07-14
[atspi-proxies] Remove alien methods from Application interface

luukvanderduim:
This removes two methods from `Application` interface that do not belong there. These belong to- and are found in `Registry` interface.

```rust
/// DeregisterEventListener method
fn deregister_event_listener(&self, event: &str) -> zbus::Result<()>;

/// RegisterEventListener method
fn register_event_listener(&self, event: &str) -> zbus::Result<()>;
```


TTWNO:
Thank you Luuk for cleaning up some of my messes!

Issue #26, opened 2022-10-03
Document `atspi`


TTWNO:
I've turned on `#[deny(missing_docs)]` this morning on `atspi` and oh boy gee golly does it have a lot of missing documentation.
I'm self-assigning this issue, and should have it documented by the end of the week.

TTWNO:
https://github.com/odilia-app/atspi/pull/47

TTWNO:
End of the week, hey?

So one year later and atill not a ton of progress here. I wish we could set it up like codecov where all new items beed to be at least as documented as the rest of the crate.

Alas, no dice. I'm going to start with adding a CI job that checks that everything is documented ("_check_doc_completion" feature flag(, and then at least we'll be aware that this is an issue and can check each time if the number of undocumented items has gone up or down.

luukvanderduim:
I care about documentation but this issue is too broad and endless. Absolutely `atspi` deserves to be documented but this does not define when the task starts nor when it is done.  
How would you feel about closing this one in favor of  opening smaller, feasible documentation issues.?
e.g. 'Document `EditableText` methods' or 'Document the `DocumentEvents` user facing event types.

TTWNO:
Great plan!

PR #116, opened 2023-07-19
[atspi-proxies] Document `AccessibleProxy`

luukvanderduim:
This adds documentation for `AccessibleProxy`

Addresses #26


luukvanderduim:
To be precise, these are from the XML protocol descriptions.

PR #117, opened 2023-07-25
Add string varaint inside UnknownBusSignature error

TTWNO:


atspi-common/src/events/mod.rs
```diff
@@ -545,7 +545,7 @@ impl TryFrom<&zbus::Message> for Event {
 		// Therefore no outer parentheses.
 		match body_signature {
 			// Marshalled Accessible signature
-			"so" => match member_str {
+			"so" | "(so)" => match member_str {
```
luukvanderduim, line 548: The reason these show up on the bus with parentheses is probably because the private implementations of the  `Cache` interface - I tried thunderbird and firefox - do their own marshalling and understood the DBus specifications a little bit differently.
Luckily the fix is super-easy. Well spotted Tait!

luukvanderduim (approved):

PR #120, opened 2023-07-27
[atspi-proxies] Document Registry's `Application::get_children`

luukvanderduim:
Document how to obtain list of accessible bus applications.
    
The implementation of `Accessible::get_children` on `org.a11y.atspi.Registry`
returns a list of root objects corresponding to all accessible applications
on the bus.


TTWNO (approved):
This is good information to be aware of. Thank you for finding and adding documentation for it.

PR #121, opened 2023-07-27
[atspi-proxies] Don't include what does not exist

luukvanderduim:
The include directive in `Cargo.toml` describes what is stored on crates' repository.
In general we only want to include what is necessary to compile the package, for all other requirements people will seek out the repository. 

The "include" directive referred to "atspi-macros" which no longer exists.

Note:

We might want to follow this up with includes for the other crates in the workspace.
We also should look into whether we want this in the top-level `Cargo.toml` of the workspace.


TTWNO:
Since crates.io handles each crate individually, I would prefer that the `include` directive is per crate.

Would you mind adding the includes to the other crates?

TTWNO:
Thanks, Luuk!

PR #123, opened 2023-07-27
Re-export zbus for convenience

luukvanderduim:
A bit of  janitorial and quality of life:

Removes export of no longer existing `atspi-client` 
Re-exports `zbus` for convenience.

### DIFF
--- a/atspi/src/lib.rs
+++ b/atspi/src/lib.rs
@@ -12,5 +12,5 @@ pub use atspi_proxies as proxy;
 #[cfg(feature = "connection")]
 pub use atspi_connection as connection;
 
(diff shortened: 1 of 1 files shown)


TTWNO (approved):
I like this change. It seems to be very common to re-export underlying libraries.

luukvanderduim:
Especially if you know (almost) all users will need it anyway its a very simple quality of life thing. 
Just add `atspi` and get `zbus` for free (if you have the feature set).
Added advantage comes when experimenting and you want to try a different or patched `zbus`, with this in, you only need to change `atspi`'s dependency and not worry about doing this in both places.

PR #129, opened 2023-08-02
Rename accessible

### LINKED ISSUE #126: Consider renaming `Accessible` to `ObjectReference`
"Accessible" does not really make sense as a name. In [Federico's GAUDEC talk](https://www.youtube.com/live/WVWrllJQJ_s?feature=share&t=4493) he shows an example of a complex structure and names these all "ObjectReference", which is much more clear and useful to newcomers and still paints the right idea.

This also allows us to get rid of the distinction between an object that implements accessible "Accessible which implements the accessible interface", and then we could write the unambiguous "object which implements the accessible interface".

TTWNO:
Closes #126


atspi-common/src/object_reference.rs
```diff
@@ -8,17 +8,17 @@ pub const ACCESSIBLE_PAIR_SIGNATURE: Signature<'_> = Signature::from_static_str_
 // check where the lifetimes of the borrow are tied to, see also: comment on `interface()` method
 // in `DefaultEvent` impl
 // then rename into Owned for this one.
-/// Owned Accessible type
+/// Owned `ObjectReference` type
 /// Emitted by `CacheRemove` and `Available`
 #[derive(Debug, Clone, Serialize, Deserialize, Type, PartialEq, Eq, Hash)]
-pub struct Accessible {
+pub struct ObjectReference {
```
DataTriny, line 14: Unless it was previously ruled out, I'd suggest renaming this struct `ObjectRef`. The name is quite long for a type so pervasive, and the `Ref` keyword should be known by most Rust developers.

DataTriny:
I would have also renamed `object_reference.rs` in the process. Otherwise changes look OK to me.

atspi-common/src/events/mod.rs
```diff
@@ -152,7 +151,7 @@ pub struct EventBodyOwned {
 
 impl From<EventBodyQT> for EventBodyOwned {
 	fn from(body: EventBodyQT) -> Self {
-		let accessible = Accessible { name: body.properties.name, path: body.properties.path };
+		let accessible = ObjectRef { name: body.properties.name, path: body.properties.path };
```
DataTriny, line 154:
```suggestion
		let object = ObjectRef { name: body.properties.name, path: body.properties.path };
```

DataTriny (changes requested):

TTWNO:
Don't worry about the coverage failing; it just happens sometimes.

luukvanderduim:
> Don't worry about the coverage failing; it just happens sometimes.

I never worry about that. I am even unsure if that metric is even all that meaningful without knowing anything about the tests in use.
I do worry somewhat about the false sense of security a green lit '90% code-coverage' might give.

DataTriny (approved):

PR #135, opened 2023-09-16
Add test for unarshalled `CacheItem` bodies.

luukvanderduim:
It appears to be common practice for bus servers to send `AddAccessible` and  `RemoveAccessible` signals with unmarshalled bodies.

This behaviour requires us to special-case parsing `Cache` events in
`TryFrom<&Message> for Event` by also checking for bodies with outer parentheses.
    
This adds a test that sends such a compensating `CacheItem` body.

This change is on top of the fix to the parser and intends to help guard we do not break `Cache` parsing (again).


TTWNO:
Thank you for adding the marshalled/unmarshalled tests.

This will bring great stability to the codebase.

PR #139, opened 2023-09-18
Remove XML Bus file

TTWNO:


luukvanderduim:
Did we rely on the service's introspection or is the launcher's description published somewhere.. 
(I cannot seem to find it in `at-spi2-core/xml` nor in `at-spi2-core/bus/` )

As I do not expect it to change any time soon, for now it is probably fine not to have it.

luukvanderduim (approved):

PR #140, opened 2023-09-18
Quicken CI

TTWNO:


DataTriny:
FWIW in AccessKit CI we use [dtolnay/install](https://github.com/dtolnay/install): it is probably as fast, but definitely more readable. And it would be able to install all the binaries you need.

Also, I really don't see the benefits of running clippy for stable and beta. Why not running it on nightly only like everyone else does these days?

TTWNO:
> Also, I really don't see the benefits of running clippy for stable and beta. Why not running it on nightly only like everyone else does these days?

I didn't know it was common to do that. I don't see any reasin why not, as long as nightly is stricter. The concern is that I don't want to be chasing down lints that end up not getting added to stable anyways.

TTWNO:
> FWIW in AccessKit CI we use [dtolnay/install](https://github.com/dtolnay/install): it is probably as fast, but definitely more readable. And it would be able to install all the binaries you need.

I'll give this a look in my PR and see if that works just as easily. Thanks for the tip!

DataTriny:
> I don't want to be chasing down lints that end up not getting added to stable anyways.

I think it never happened to us once, and we have set up our CI more than two years ago now. clippy probably reached some sort of maturity at this point.

Issue #125, opened 2023-07-27
Match signatures with XML with tests


TTWNO:
As [Federico mentions in his GAUDEC talk](https://www.youtube.com/live/WVWrllJQJ_s?feature=share&t=4581) we should be matching the signature of our various types directly with the XML.

```rust
assert_eq!(
    MyType::signature(),
    signature_from("Cache.xml", "MySignal"),
);
```

luukvanderduim:
Working on it.

PR #141, opened 2023-09-26
Type validation with zbus_lockstep!

luukvanderduim:
This adds [zbus-lockstep](https://github.com/luukvanderduim/zbus-lockstep/) type validation to _atspi_.

Addresses:  #125

This replaces, in many cases, comparing signatures against a hard-coded const.
Also this adds a markdown page with an overview of types which have and have received lockstep validation.
(Some types simply can't because we have chosen a more efficient data-type.)

This is a good start, there are more validations to be added but the existing types with such tests have been converted:

`atspi` type| XML validation | Notes
|:--|---|---|
| `ObjectRef`| ✓ | Formerly: `Accessible`, used in `Available` and `RemoveAccessible`
| `CacheItem`| ✓ | See: `AddAccessible` signal of the `Cache` interface.
| `LegacyCacheItem` | ⚠ | The signature does not appear in current protocol: test hard coded.
| `Interface` | ⚠ | Unlike the protocol, which uses strings, we store `Interface` as bitflags.
| `InterfaceSet` | ⚠ | We store `InterfaceSet` as a single `u64`, not as 'array of strings'.
| `State` | ⚠ | XML: "au", a list of enum variants. Whereas `atspi` stores a `u64` bitflag.
| `StateSet` | ⚠ | See `State`.
| `EventBodyOwned` | ✓ | Covers many other events.
| `EventBodyQt` | ⚠ | Signature does not appear in XML.
| `EventListeners` | ✓ | Represents signals of `EventListenerRegistered` and `EventListenerDeregistered`
| `SortOrder` | ✓ | 'sortby' argument pf member `GetMatches`
| `TreeTraversalType` | ✓ | 'tree' argument of member `GetMatchesTo`
| `MatchType` | ✓ | Found at index 3, in the 'rule' argument in `GetMatches`
| `CoordType` | ✓ | Argument of `GetImagePosition` of `org.a11y.atspi.Image`
| `ClipType` | ✓ | `Text::GetTextAtOffset` argument `type`
| `Granularity` | ✓ | `Text::GetStringAtOffset` argument `granularity`
| `Layer` | ✓ | `Component::GetLayer` return type.
| `ScrollType` | ✓ | `Component::ScrollTo` argument `type`
| `Live` | ✓ | `Event.Object::Announcement` argument `politeness`

Note: 
 that the `const` signatures still have a purpose in run-time code because the lockstep functions are not `const` themselves. When those can be converted to `const fn` we may be able to leverage const  evaluation on those `const`s too.

## Caveat

This will probably add to test duration because the tests now regularly visit XML files to lookup signatures.

## Future 

It might be an improvement to add a function to `zbus_lockstep` to, if updates are available, retrieve the protocol descriptions.
This function could then be added to a `build.rs`, needs a bit more thinking.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,7 @@ jobs:
           toolchain: stable
       - name: Install Cargo Machete
         uses: taiki-e/install-action@cargo-machete
-      - name: Check For Unused Depdencies
+      - name: Check For Unused Dependencies
         run: cargo machete
   semver-compliance:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,7 +194,7 @@ jobs:
         with:
           tool: cargo-semver-checks
       - name: Check Semver Compliance
-        run: cargo semver-checks check-release -p atspi --default-features
+        run: LOCKSTEP_XML_PATH="../xml" cargo semver-checks check-release -p atspi --default-features
   msrv-compliance:
     runs-on: ubuntu-latest
     needs: [clippy,no-unused-dependencies,find-msrv]
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Part of the [Odilia screen reader project](https://odilia.app).
 ## Design
 
 * Fully documented, with `#[deny(missing_docs)]`
-	* Or at least, it will be by 1.0
+* Or at least, it will be by 1.0
 * Fully safe, with `#[deny(unsafe_code)]`
 * Fantastic code style with `#[deny(clippy:all, clippy::pedantic, clippy::cargo)]`
 
@@ -24,6 +24,17 @@ We use the asynchronous zbus API, so to use atspi, you will need to run an async
 [tokio](https://crates.io/crates/tokio) or
 [async-std](https://crates.io/crates/async-std).
 The `async-io` and `tokio` features are exposed and will be passed through to zbus.
+
+## D-Bus type validation
+
+Atspi is used to send and receive data to and from applications. Sender and recipient need to agree on the shape of the data type for fruitful communication. Our best bet is to keep our types in sync with the protocol descriptions.
+
+We employ [zbus-lockstep](https://github.com/luukvanderduim/zbus-lockstep/) to match types against those defined in the AT-SPI2 protocol descriptions.
+
+Not all types can be validated (easily) with zbus_lockstep because types may not exist
+in the protocol descriptions, for example because they are deprecated (but still in use) or we have chosen a different representation.
+
+[A (partial) review of type validation may be found here](type_validation.md)
 
 ## License
 
--- a/atspi-common/Cargo.toml
+++ b/atspi-common/Cargo.toml
@@ -27,6 +27,7 @@ zvariant = { version = "4.0", default-features = false }
 zbus = { workspace = true, optional = true, default-features = false }
 zbus-lockstep = "0.4.1"
 zbus-lockstep-macros = "0.4.1"
+zbus_xml = "4.0.0"
 
 [dev-dependencies]
 zbus = { workspace = true }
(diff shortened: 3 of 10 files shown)


luukvanderduim:
Wrt to the semver CI err, it seems the runner does something odd with the tree? 
Need to think this one over on how to fix this properly. Bah.

TTWNO:
Looks like semver is complaining sbout missing XML files.... My guess is it's running `cargo doc` within `atspi-common` instead of from fhe root, where we'd typically expect it to run.

But I also haven't gone over the PR in excrutiating detail yet. I'll have time tomorrow afternoon to look at all the changes, and then maybe fix the semver issue.

luukvanderduim:
> Looks like semver is complaining sbout missing XML files.... My guess is it's running `cargo doc` within `atspi-common` instead of from fhe root, where we'd typically expect it to run.

That is probably what happens.
Without path argument, the attribute macro looks for the default xml directory in the current directory, either `XML` or `xml`. 
I want to replace that with a recursive search from workspace- or crate root which is likely more robust. However the only method of finding the workspace root path I found up until this point was by parsing output from running 'cargo'. I thought that not so elegant - but maybe it is the only way. 

> 
> But I also haven't gone over the PR in excrutiating detail yet. I'll have time tomorrow afternoon to look at all the changes, and then maybe fix the semver issue.

Thanks!

TTWNO:
This seems to break the WASM compilation target due to the inclusion of `zbus` itself, instead of its sub-components: `zvariant`, `zbus_names`, etc.

I realize you need the XML definitions... Would you consider getting `zbus` to publish their zbus_xml` crate so you can include that and drop the dependency on the `zbus` crate?

I realize it's a big ask, since this is a library you maintain yourself. I'd be willing to help if you want :)

TTWNO:
The good news is that thanks to your environment variable override, I got the `semver-checks` step to pass! Great foresight to see that some developer would eventually need it.

luukvanderduim:
> The good news is that thanks to your environment variable override, I got the `semver-checks` step to pass! Great foresight to see that some developer would eventually need it.

I needed it myself in CI tests, I believe. ;) It was trivial to add.

luukvanderduim:
> This seems to break the WASM compilation target due to the inclusion of `zbus` itself, instead of its sub-components: `zvariant`, `zbus_names`, etc.
> 
> I realize you need the XML definitions... Would you consider getting `zbus` to publish their `zbus_xml` crate so you can include that and drop the dependency on the `zbus` crate?
> 
> I realize it's a big ask, since this is a library you maintain yourself. I'd be willing to help if you want :)

Not that big of an ask since [subcrate `zbus-xml` is underway](https://github.com/dbus2/zbus/commits/main/zbus_xml). We are awaiting `zbus 4.0` which will bring this too.

TTWNO:
Fantastic! Looking forward to it!

TTWNO:
Looks like zbus4.0 comes with some fantastic features:

* [generated traits for proxies](https://github.com/dbus2/zbus/issues/433)
* [drop Arc<T> wrappers for Messages](https://github.com/dbus2/zbus/commit/1856c0f348eff0d25fac0921c963cb05b3811c59)
* [fix large futures warning](https://github.com/dbus2/zbus/pull/458) that I filed back when they were on Gitlab
* [and more](https://github.com/dbus2/zbus/milestone/3)

Very much looking forward to it!

luukvanderduim:
We do. I've put it there temporarily. It looks like the macro expands to code that requires zbus_xml to be in scope.
I'm fixing this in zbus_lockstep - in the meantime I am a bit puzzled about things rust-analyzer says..

luukvanderduim:
`zbus-lockstep` received an update such that the convenience macros no longer reuire `zbus_xml` to be in scope.

I did not update `ci.yaml` - which previously required the env variable `LOCKSTEP_XML_PATH` to be set. 

The semver compliance check still appears to need it. Let me try to fix that.

TTWNO:
Great work, Luuk! I'll review this one this evening.

PR #143, opened 2023-10-09
Document `Action` iface

luukvanderduim:
This series documents `ActionProxy`, the handle pertaining to `org.a11y.atspi.Action`.


TTWNO:
Damn but it bring codecov down to 89%. Guess we can't merge that, lol!

PR #145, opened 2023-10-10
Clippy items after test mod

luukvanderduim:
Addresses a Clippy warning where items are placed after test module.


TTWNO:
Although it seems that this lint was added in 1.71, I can only seem to trigger it on nightly.

There's no harm in merging this, I just want to understand why this lint is so hard to trigger...

luukvanderduim:
I just fix the squiglies sir. ;)

I saw it popup and I do agree with the suggestion. For me at least it is unexpected to find items after a `test` module.

It probably only pops up if you lint for 'Style' group lints. 

[items_after_test_module lint](https://rust-lang.github.io/rust-clippy/master/index.html#/items_after_test_module?groups=style)

PR #147, opened 2023-10-10
Idiomatic constructors for `AccessibilityConnection`

luukvanderduim:
[connection] Rename `open` and `connect`
    
This proposes to rename `open` to `new`, and `connect` to `from_address` as this
follows the conventional naming around constructing instances:
    
[constructor guideline](https://rust-lang.github.io/api-guidelines/predictability.html?highlight=new#constructors-are-static-inherent-methods-c-ctor)
    
This also adjusts tests and examples accordingly.
    
Note this is a breaking change. That being said, the breakage is simple in
nature.


luukvanderduim:
Bah, this unintentionally included the 'Fix items_after_test_module' commits too, let me revert those.

PR #148, opened 2023-10-20
Add Conditions to CI

TTWNO:


TTWNO:
This slightly increases test time, but reduces the issue of running huge jobs like `tests` if `clippy` or `rustfmt` doesn't even pass (since those take so little time).

This would trade parallelism for efficiency, somewhat. I don't want to run them all sequentially, since most tests usually do pass, but it puts up a low barrier to stop obviously bad commits from being checked thoroughly.

TTWNO:
I will need to make more changes to this though, since `cargo machete` seems to run in 7 seconds, which would also be a great barrier job.

Feel free to shut me down if I'm being silly.

PR #150, opened 2024-01-12
Clippy fixes jan24

luukvanderduim:
Clippy and I had fun today.

### DIFF
--- a/atspi-common/src/events/mod.rs
+++ b/atspi-common/src/events/mod.rs
@@ -784,7 +784,7 @@ mod tests {
 		let accessible = crate::ObjectRef::default();
 		let name = accessible.name;
 		let path = accessible.path;
-		let props = HashMap::from([(name, ObjectPath::try_from(path).unwrap().into())]);
+		let props = HashMap::from([(name, ObjectPath::from(path).into())]);
 		assert_eq!(event_body.properties, props);
 	}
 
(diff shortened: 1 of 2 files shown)


TTWNO:
Curious... why didn't this show up on CI before?

Were these unstable lints that just got moved into stable?

TTWNO:
Also, good work! I always appreciate a good, simple, cleanup PR.

TTWNO (approved):
Much cleaner and easier to read, throughout!

luukvanderduim:
> Curious... why didn't this show up on CI before?
> 
> Were these unstable lints that just got moved into stable?

Curious indeed.

`clippy::unnecessary_fallible_conversions` was added in 1.75.0, so this one is new in stable.
`clippy::get_first` was added in 1.63.0, which is quite a bit older.

Both are `style`-class warnings which are, strictly speaking, not wrong or unsound, but there just is a better alternative.
Maybe CI was not checking for these earlier?

I had seen the `get_first()` warning before locally, but decided against fixing it because we would end up with `get_first()` on the left hand side and `get(1)` on the right hand side - which I found an ugly imbalance. However, I realized, in these cases where we get `Value::Structure(s)` fields slice, containing the fields of an `ObjectRef`. We assert its signature first, so we are guaranteed to get a slice of two fields.  Then `get_first()` and `get_last()` are nicer.

I do not understand why CI did not pick up on `clippy::get_first` before.

luukvanderduim:
The lint had a bug it seems, which mad [it not warn on non-primitive types](https://github.com/rust-lang/rust-clippy/issues/11594)

No longer a mystery!

PR #153, opened 2024-02-07
[bench] Add criterion parse `Message`s benchmark

luukvanderduim:
This adds [Criterion](https://bheisler.github.io/criterion.rs/book/) based benchmarking.

```console
cargo bench --bench atspi_benches
```
Will run the benchmark.
The benchmark generates a `Vec` of 1000 random `Message`s and will time how long it takes to parse these into user facing `Event` types.

Note that running 
```console 
cargo bench
``` 
without specifying the benchmark will run all unit tests too.

This pull-request is created as follow-up on the dependency fixups.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,27 @@ on:
     branches:
       - main
 jobs:
+  benchmarks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v3
+      - name: Cache cargo home
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-cargo-home
+        with:
+          path: |
+            ~/.cargo/bin
+            ~/.cargo/registry/index
+            ~/.cargo/registry/cache
+            ~/.cargo/git/db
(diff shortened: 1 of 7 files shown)


TTWNO:
Could you please add a second test for an additional order of magnitude (100,000)?

I know we've tested Thunderbird as giving nearly... what was it, 80 thousand events in one second?

It'd be good to know how `atspi` performs at that scale.

TTWNO:
> At current precision this would add considerable time to CI.

I'm not going to run the benchmark in CI... Just make sure that it compiles :)


> Since the numbers scale perfectly linear, the greater test adds little to our understanding of 'parsing efficiency'.

Fair point. Glad to hear it is almost exactly linear.

> Should you want to stress-test atspi in more real-world fashion, we may want to do something else. But that is arguably outside the scope of this CI as the efficiency of zbus and DBus are not in our hands. Interesting, but not in CI.

Agreed.

luukvanderduim:
We now have two separate benchmarks:

'event_parsing' and 'event_parsing_100k'. 
Hopefully this is what you wanted @TTWNO ?

Issue #154, opened 2024-02-12
port to zbus 4.0


TTWNO:
Given that `zbus` has released version 4, it should probably be upgraded to.

If not only for the fact that it is a dependent for #141

TTWNO:
@luukvanderduim would you like to take on this one?

I don't want to duplicate work.

luukvanderduim:
Yes, just remove this branch, I am working on the port as we speak.

PR #155, opened 2024-02-16
zbus 4.0 port

TTWNO:
Good work so far, Luuk!

I'm just opening this so we can catalogue some of the issues we've had porting.

( Addresses #154 )


DataTriny:
Hello there,

We should of course upgrade to zbus 4 at some point, but now is way too soon for AccessKit. We just bumped our MSRV to 1.68 so I wouldn't want to be stuck on the current version of this crate for a while because of that.

DataTriny:
@TTWNO It all depends on when this will be ready and if I find things to add or fix here in the meantime.

I guess we'll all have to agree on how and when to bump our MSRV in sync at some point.

luukvanderduim:
I understand you cannot bump MSVR now, however we do want to move on. 
Zbus 4 offers a number of improvements we would like to leverage / benefit from.

For the time being `AccessKit` could depend on the `atspi` version you are currently on.
I believe `AccessKit` depends on `atspi` (mostly) for types doesn't it? 

I have been pondering if a types-and-traits crate in the workspace could be something worth exploring.

TTWNO:
> I have been pondering if a types-and-traits crate in the workspace could be something worth exploring.

Would this just be `atspi-common`?

luukvanderduim:
> > I have been pondering if a types-and-traits crate in the workspace could be something worth exploring.
> 
> Would this just be `atspi-common`?

That would make things easier if we could..

Let's see:
-  `zbus` is an optional dependency on `atspi-common`, so that may work.
- However, since `zbus` is a dependency (albeit being optional) we do have an MSRV.
- Maybe if we remove the 'rust-version' key in this crate's `Cargo.toml` we could avoid forcing downstream users of the types into a new(er) rust version - as long as they  do not enable `zbus`?

Then we'd need to carefully look at which types and traits we need to move over still.

TTWNO:
> Then we'd need to carefully look at which types and traits we need to move over still.

This sounds like more things to keep straight in our heads. Personally, I don't think that's a good idea.

atspi-common/src/events/mod.rs
```diff
@@ -131,7 +103,7 @@ impl Default for EventBodyQT {
 /// Standard event body (GTK, `egui`, etc.)
 /// NOTE: Qt has its own signature: [`EventBodyQT`].
 /// Signature `(siiva{sv})`,
-#[derive(Clone, Debug, Serialize, Deserialize, Type, PartialEq)]
+#[derive(Debug, Serialize, Deserialize, Type, PartialEq)]
```
TTWNO, line 106: Why was the `Clone` derive removed here?

atspi-common/src/events/mod.rs
```diff
@@ -170,12 +152,41 @@ impl Default for EventBodyOwned {
...
-			any_data: Value::U8(0u8).into(),
+			any_data: 0u8.into(),
 			properties: HashMap::new(),
 		}
 	}
 }
 
+// Manual implementation to account for the remote chance that either `any_data` or `properties`
+// fail to clone because the clone would exceed the open file limit.
+impl Clone for EventBodyOwned {
```
TTWNO, line 163: Oh I see... Makes sense now.

atspi-common/src/events/mod.rs
```diff
@@ -170,12 +152,41 @@ impl Default for EventBodyOwned {
...
+		Self {
+			kind: self.kind.clone(),
+			detail1: self.detail1,
+			detail2: self.detail2,
+			any_data: cloned_any_data,
+			properties: cloned_properties,
+		}
+	}
+}
+
```
TTWNO, line 189: Should we just let the users of atspi call `.try_clone()` instead? And let them decide whether or not to fail?

atspi-common/src/events/mod.rs
```diff
@@ -442,11 +448,14 @@ impl TryFrom<&zbus::Message> for EventBodyOwned {
...
-		} else if signatures_are_eq(&signature, &ATSPI_EVENT_SIGNATURE) {
-			Ok(message.body::<EventBodyOwned>()?)
+		let body = message.body();
+		let signature = body.signature().ok_or_else(|| AtspiError::MissingSignature)?;
+
+		if signature == QSPI_EVENT_SIGNATURE {
+			let qt_body = body.deserialize::<EventBodyQT>()?;
+			Ok(EventBodyOwned::from(qt_body))
+		} else if signature == ATSPI_EVENT_SIGNATURE {
+			Ok(body.deserialize::<EventBodyOwned>()?)
```
TTWNO, line 458: Please add a test so that all three cases are covered (QT is currently not).

atspi-common/src/events/mod.rs
```diff
@@ -787,37 +798,4 @@ mod tests {
...
-		assert_eq_signatures!(without_parentheses, without_parentheses);
-		// test false cases with unbalanced parentheses
-		let with_parentheses = &Signature::from_static_str_unchecked("(ii)(ii)");
-		let without_parentheses = &Signature::from_static_str_unchecked("((ii)(ii)");
-		assert!(!signatures_are_eq(with_parentheses, without_parentheses));
-		// test case with more than one extra outer parentheses
-		let with_parentheses = &Signature::from_static_str_unchecked("((ii)(ii))");
-		let without_parentheses = &Signature::from_static_str_unchecked("((((ii)(ii))))");
-		assert!(!signatures_are_eq(with_parentheses, without_parentheses));
-	}
```
TTWNO, line 822: Glad to see this logic upstreamed.

atspi-common/src/events/object.rs
```diff
@@ -107,9 +107,41 @@ pub enum Property {
...
+			}
+			Property::TableRowHeader(table_row_header) => {
+				Self::TableRowHeader(table_row_header.clone())
+			}
+			Property::TableSummary(table_summary) => Self::TableSummary(table_summary.clone()),
+			Property::Other((property, value)) => Self::Other((
+				property.clone(),
+				value.try_clone().unwrap_or_else(|err| {
+					eprintln!("Failed to clone value: {err:?}");
+					u8::default().into()
```
TTWNO, line 135: Again, I would argue that we should give the `.try_clone()` option to users and let them handle errors.

On the other hand, it's not a trait (in `std` anyways), and it is exceedingly rare that this would actually be triggered.... Hmmmm

Does cloning a file descriptor *do* anything? Wouldn't the "limit" only be exceeded if we tried to open it?

atspi-proxies/Cargo.toml
```diff
@@ -9,7 +9,7 @@ homepage = "https://github.com/odilia-app/atspi"
 keywords = ["screen-reader", "accessibility", "a11y", "tts", "linux"]
 categories = ["accessibility", "api-bindings"]
 edition = "2021"
-rust-version = "1.68.2"
+rust-version.workspace = true
```
TTWNO, line 12: Although this is valid... It breaks the `cargo msrv verify` command. Specifying a crate like `atspi` (the crate, not the workspace), or attempting to run it in the workspace: neither work.

Although I agree we should probably just have a workspace MSRV, I'm wondering if there's a better verification tool that will handle the `.workspace` suffix?

DataTriny:
@TTWNO Do you need `cargo msrv` in the first place?

I recently enforced MSRV for AccessKit:

- I defined the Rust version once in the workspace manifest,
- I made all members inherit the property: `rust-version.workspace = true`,
- In the CI I used `sed` to grab the version defined on the workspace: `echo "version=`cat Cargo.toml | sed -n 's/rust-version = "\(.*\)"/\1/p'`" >> "$GITHUB_OUTPUT"`,
- And in the test jobs I installed Rust, specifying the version:

```yaml
      - name: install stable toolchain
        uses: dtolnay/rust-toolchain@master
        with:
          toolchain: ${{ needs.find-msrv.outputs.version }}
```

If your CI fail with compile errors, then you know you broke your MSRV policy. It's usually enough since you also have a clippy job able to catch regular compile errors, should there be some. You might lose a bit of efficiency since you won't reuse the cache of the clippy job in the test jobs, not a big deal IMO.

zeenix:
> I guess we'll all have to agree on how and when to bump our MSRV in sync at some point.

I think it would be a good idea to have some (even vague) policy about it. Just for the record, I bumped MSRV in zbus to latest of the time only because it was a major version bump. I will be a lot more reluctant to bump it going forward until the next API break.

TTWNO:
@DataTriny great idea! I'll try that today and commit it.

atspi-common/src/events/mod.rs
luukvanderduim: `EventBodyOwned` has fields of type `OwnedValue`. 
In zvariant 4, `Value`s no longer have `Clone` because of the [safe Fd work](https://github.com/dbus2/zbus/pull/395). Cloning is now achieved through [`Value::try_clone`](https://docs.rs/zvariant/latest/zvariant/enum.Value.html#method.try_clone).

I manually re-implemented `Clone` using `OwnedValue::try_clone` to not break our API, because: 
- Suddenly missing `Clone` is pretty rough.
- There is only a remote chance that we will hit the error path in this context, because it requires the current process to exceed the maximum number of open files: (from `Value::try_clone` docs)

```text
Errors

This method can currently only fail on Unix platforms for [Value::Fd](https://docs.rs/zvariant/latest/zvariant/enum.Value.html#variant.Fd) variant containing an [Fd::Owned](https://docs.rs/zvariant/latest/zvariant/enum.Fd.html#variant.Owned) variant. This happens when the current process exceeds the maximum number of open file descriptors.
```

atspi-common/src/events/mod.rs
luukvanderduim: I decided to be 'lust practical' considering the context:

Did or will `EventBodyOwned.any_data` ever contain a file?
Did or will `EventBodyOwned.properties` ever contain a file?

I don't think so and I think the chances are slim they ever will and should they ever, the chances of triggering the failure path is slim.

It expected that a type has `Clone`.

atspi-common/src/events/object.rs
luukvanderduim: AFAIU, having the `OwnedFd` is `having an open file`, otherwise cloning these would not be limited?

I fully agree that the chances are slim. Have we ever seen file descriptors being sent? 
If we can, I would like to not burden our users with `try_clone()` but just keep a `Clone` implementation.

atspi-proxies/Cargo.toml
luukvanderduim: > I'm wondering if there's a better verification tool that will handle the `.workspace` suffix?

If it makes our lives easier, I'll happily revert this change.

luukvanderduim:
> I think it would be a good idea to have some (even vague) policy about it. Just for the record, I bumped MSRV in zbus to latest of the time only because it was a major version bump. I will be a lot more reluctant to bump it going forward until the next API break.

Would you by any chance have something in mind?

There is of course versioning policy: Considering we are <v1.0.0 and `atspi` is still a mess, breakage is to be expected.
That being said, I am open to catering `AccessKit`'s needs by moving the AT-SPI2 types into a sub-crate without msrv.

zeenix:
> @zeenix Do you have something in mind?

It depends entirely on the maintainers according to the project's needs and I don't have enough context to suggest something specific. My suggestion was only that there should be policy (even if vague). In zbus, I'd try my best to not bump MSRV to anything newer than 6 months in a non-major release.

DataTriny:
@luukvanderduim `atspi-common` would still have to depend on `zvariant` and `zvariant-derive`. But in AccessKit we use some proxies from `atspi-proxies` to grab the bus address and talk to the registry anyway. So it wouldn't help much.

I'd be curious to know what features you absolutely need from `zbus` 4.0 right now.

luukvanderduim:
> @luukvanderduim `atspi-common` would still have to depend on `zvariant` and `zvariant-derive`. But in AccessKit we use some proxies from `atspi-proxies` to grab the bus address and talk to the registry anyway. So it wouldn't help much.

You would be using `atspi-connection` for the bus address and `atspi-proxies` for the registry proxy. Nice to hear you're using those. 

> I'd be curious to know what features you absolutely need from `zbus` 4.0 right now.

- We have been waiting to adopt type validation against AT-SPI2 protocol descriptions using [zbus-lockstep](https://github.com/luukvanderduim/zbus-lockstep) - for which zbus 4 is needed. See: #141 
- We have upstreamed a `Signature` custom `PartialEq::eq`, which allows atspi to lose weight.
- Moving to zbus 4.x yields [a decent performance uptick of ~3x in `Message` -> `Event` parsing](https://gist.github.com/luukvanderduim/e874c446e4a206a4aa523c80424f1a8f)

atspi-proxies/Cargo.toml
TTWNO: After further explanation from DataTriny, this looks like an easily solvable problem.

TTWNO:
@luukvanderduim 

Ok, I'm coming around to your point on the `Clone` implementation. I've added a safety comment locally that links to `zvariant::OwnedValue` for more info.

However, I see there are a couple places where if you can not `try_clone()` correctly, instead of `panic`ing, you use a default value:

- `atspi-common/src/events/object.rs:135`
- `atspi-common/src/events/object.rs:235`

Should these not also panic? I want to have consistent behaviour for library users. Even if that does mean that it can technically panic in more places.

TTWNO:
Or, would you argue, like in `./atspi-common/src/events/object.rs:1239` that the comment is simply unnecessary due to the remoteness of the issue? Sorry if we're covering old ground, it's late here.

TTWNO:
@DataTriny I've released atspi `v0.21.0` with the latest changes (pre merging this branch). I've kept the same MSRV, dependencies (I think we even removed one), and especially `zbus` version.

Any future additions to the crate will be gated behind `zbus` 4.x.

TTWNO:
> All of this is very unfortunate...

I don't see the problem here? accesskit is often 6 months or so behind the latest version anyways; is there a reason you need us to keep compatibility with zbus 3.x for that long?

Or is there some other issue I'm not seeing here?

zeenix:
Maybe the README is outdated but given the beta status of this project, I don't really see why bumping MSRV is a big problem here. 🤔

DataTriny:
I just don't want to propose bumping AccessKit's MSRV to 1.75 before our biggest downstream users bring the subject, namely egui (currently on Rust 1.72) and slint (currently on Rust 1.70).

TTWNO:
> I just don't want to propose bumping AccessKit's MSRV to 1.75 before our biggest downstream users bring the subject,

Then don't. It's ok. Like I said, I often see accesskit being many months behind the latest atspi version. Accesskit's dependence on us has been really mostly the most fundemental types. Those aren't likely to change in the near futute.

I see no problem bumping this project, even if accesskit is not ready to.

PR #158, opened 2024-02-29
Remove expect formatting

TTWNO:


luukvanderduim:
Fixed expect messages to reflect why we think these will succeed - if we think so.

Also added some clarification and docs in macros.rs because it was
not immediately apparent what was going on.

In the "Should fail" tests I had long winded explanations too, but decided to
revert those for this PR.

In tests.rs we panic'ed on a method `sender(unique_name)` that can only fail if
the provided name cannot be converted into a `UniqueName<'a>`.
Since we provide it with a valid, owned variant, this will always succeed. 

Which is why I removed your `unwrap_or_else(|err| panic!("bla with {err}"))` in those instances.

TTWNO:
Thanks again, Luuk! I rebased the changes on latest main, and these are much better and make more sense. I understand the semantics around expect much better now, and will try to use it right in the future.

PR #161, opened 2024-03-09
Update XML from upstream

TTWNO:
Do not merge yet. I think we have a new method to add to the Socket proxy.

### DIFF
--- a/atspi-common/src/lib.rs
+++ b/atspi-common/src/lib.rs
@@ -209,7 +209,7 @@ pub enum ScrollType {
 /// The argument in the `Announcement` event is named `politeness`.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize, Type)]
 #[repr(i32)]
-pub enum Live {
+pub enum Politeness {
 	/// No live region.
 	#[default]
 	None = 0,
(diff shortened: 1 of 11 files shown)


TTWNO:
Why did this branch before your CI fix? Strange.

luukvanderduim:
It did not, I sneakily added a 'fix'.
Unfortunately it does not seem to fix anything..
The error messages looked like this: 
```console
No XML path provided and default XML path not found. Current dir: "/home/runner/work/atspi/atspi/atspi-common" 
```
Setting the environment variable to "../xml" broke things in more interesting ways. Not sure what is going on. 
It cannot find the directory and I see other problems pop-up that do not make sense at all since all other CI tests succeed.
Do you understand this better?

TTWNO:
TODO: implement `Get/Set TextSelections` for `Document`.

atspi-common/src/lib.rs
```diff
@@ -209,7 +209,7 @@ pub enum ScrollType {
 /// The argument in the `Announcement` event is named `politeness`.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize, Type)]
 #[repr(i32)]
-pub enum Live {
+pub enum Politeness {
```
luukvanderduim, line 212: Big thanks for this. 
Maybe in a certain context `Live` made sense at some point. (I suppose 'the way we may interrupt 'live' (ongoing) speech'.)
I think `Politeness` is much more descriptive.

PR #165, opened 2024-03-18
Moved `xml` definitions to `atspi-common`

luukvanderduim:
Also make sure these are around when we publish by adding them in `include` of `Cargo.toml`,

Upgrade to `zbus-lockstep` v0.4.4

Also, set tokio's default-features (instead of default_features)


atspi-common/Cargo.toml
```diff
@@ -19,14 +19,14 @@ async-std = ["zbus/async-io"]
 tokio = ["zbus/tokio"]
 
 [dependencies]
+zbus-lockstep = "0.4.4"
```
TTWNO, line 22: Sort dependencies alphabetically. No sense moving them around.

TTWNO (approved):

Issue #157, opened 2024-02-21
Remove manual implementation of `Serialize`, `Deserialize`, and `From<u32>` for `-common` types


TTWNO:
`Role` has a manual implementation of `From<u32>` and `Into<u32>` so that it can be converted to numbers easily. This should be some kind of derive macro. I noticed that we even test our implementation (because I messed it up at one point). The argument for this is that adding new ones will be easier, since the implementation's `match` block will require us to cover all cases.

Similar deal with `State`, `Interface`, they have manual `Serialide` and `Deserialize` implementations for the same reasons. Ideally, this could just be handled. IIRC, the reason for this was that I needed constant static strings of the interfaces and states, and I couldn't use a constant static string as a value for serde's de/serialize customizations.

I'll gladly close if that's still the case.

luukvanderduim:
I am not sure I fully follow your train of thought, but I'll try. Please correct me where I go wrong.

> `Role` has a manual implementation of `From<u32>` and `Into<u32>` 
 Implement `From`, get `Into` for free.

> so that it can be converted to numbers easily. This should be some kind of derive macro.
I think we can safely cast.

`Role` is a simple type:
```xml
    <method name="GetRole">
      <arg direction="out" type="u"/>
    </method>
```
It is unambiguously defined as a `u32`. Should you want the `u32` of a `Role`, casting `Role` is fine provided  `Role` has `#[repr(u32]`, which it has:

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Type, Hash)]
/// Role docs
#[repr(u32)]
pub enum Role {
```

I may have a problem with this one:
```rust
TryFrom<u32> for Role
```
which I think is unneeded. If you want a `Role` in Rust, you will just create a `Role::*`. 

I also think both attribute macros should not be separated by item docs.


> Similar deal with `State`, `Interface`, they have manual `Serialide` and `Deserialize` implementations for the same reasons. Ideally, this could just be handled. IIRC, the reason for this was that I needed constant static strings of the interfaces and states, and I couldn't use a constant static string as a value for serde's de/serialize customizations.
> 
> I'll gladly close if that's still the case.

`State` has derived `Serialize` and `Deserialize`. Let's review:

## State

`State` is a bit different, because we decided to keep it as `BitFlags`. A single state is a `u32`. 

`GetState` returns an array of `State`s which we keep as `StateSet(Bitflags<State>)`

```xml
    <method name="GetState">
[...]

PR #172, opened 2024-04-07
Enum enrichment

luukvanderduim:
This commit hopes to address (in part) #157

Adds strum and derives `Display`, `FromRepr` and `IntoStaticStr` on public enums (where this makes sense).

`FromRepr` adds `T::from_repr(U)` where `U` is the underlying type of the enum.
`FromStaticStr` implements `From<Enum> for &static str` (implicitly `Into` too).
`Display` implements `Display` (implicitly does `ToString`)

nit: I do not like how adding the derives causes the `#[derive( .. )]` span several lines.
With this PR, the `Interface`'s `Deserialize` implementation avoids copying the string, thus may have become a little bit faster for what that is worth. 

Various other changes and added tests. Please review and scrutinize before merging.


DataTriny:
@luukvanderduim I think we should first use serde in places where it is currently missing, and then use strum for what is left. Right now you have added strum as a required dependency, which is not good.

TTWNO:
> Right now you have added strum as a required dependency, which is not good.

Although I do sympathize with you wanting to keep accesskit's dependency tree small, I think this is a very small ask here. `strum` is not a particularly heavy crate, and it only depends on 3 sub crates which are not already transitive dependencies of `atspi` (one of which is simply the `strum_macros` crate itself).

Again, I'm fine restricting the use of large libraries going forward, since Odilia has a similar problem with a massive number of dependencies. But I think this is an incredibly small addition which decreases manual maintenance by a large amount: serialization is the vast majority of what `atspi-common` does.

Additionally, the new `zbus` 4.0 release has dropped a few dependencies, moving us down from 225 to 218, so we will still create a new loss across the `zbus` 4.0 transition.

luukvanderduim:
> @luukvanderduim I think we should first use serde in places where it is currently missing, and then use strum for what is left. Right now you have added strum as a required dependency, which is not good.

@DataTriny 

In the issue discussion we had earlier:

[> Since our remaining usecase cannot be solved by serde, I'd be OK having strum as an optional dependency.](https://github.com/odilia-app/atspi/issues/157#issuecomment-2001971456)

So I am a bit confused to see that you _do_ in the end wish to use serde.

DataTriny:
@luukvanderduim sorry if I wasn't clear: I was refering to an upcoming use case for which the needs are not yet fully defined, but could require the use of `strum`. However I don't think it is a good idea to address this now.

With #174 I started removing some manual implementations of the `serde` traits, and I think we should try to do it wherever we can.
The problematic instances are the event body structs. But these structs have another issue: you must know the AT-SPI internals to use them properly. We should find a way to better use the type system to represent them, which will certainly allow us to directly serialize/deserialize the types that they carry such as `Role` and `State`.

DataTriny:
@luukvanderduim 

> The StateChanged signal (see: Event.xml) uses 'string' to encode the state,
(NOT by a u32 representing the enum variant! as one might expect)
In atspi State's Rust representation is a u64 bitfield, which is then serialized /
deserialized to / from a string.

I know all of this, in fact I brought `State` and `StateSet` here from AccessKit where I had implemented them. I was precisely refering to this signal. Since `StateSet` manually implements the `serde` traits, and doesn't rely on the serialization/deserialization implementations of the `State` type to accomplish this, and since to my knowledge there is nowhere in AT-SPI where a single state is represented by its numerical representation, I think the way forward is to modify our event body types so that they can directly contain a `State` where needed. This is why I added `#[serde(rename_all = "kebab-case")]` to the enum in the first place.

> The only thing I found is that AccesKit has a specific display requirement that needed attention.

I don't know what you are refering to. Can you please point out what you've found on AccessKit? If there is something valuable to add here, I'll gladly do it.

luukvanderduim:
> @luukvanderduim
> 
> > The StateChanged signal (see: Event.xml) uses 'string' to encode the state,
> > (NOT by a u32 representing the enum variant! as one might expect)
> > In atspi State's Rust representation is a u64 bitfield, which is then serialized /
> > deserialized to / from a string.
> 
> I know all of this, 

Good. I hope you found the summary palatable.

You mentioned we get `State` and `Role` were obtained from `EventBody` , I presume you meant the `Message`'s body then.

> > The only thing I found is that AccesKit has a specific display requirement that needed attention.

I meant the scheme with spaces.
Maybe serde `rename_all` has a scheme to support lowercase-with-spaces, strum, afaik does not.

```Rust
        /// Object is a label indicating the keyboard accelerators for the parent.
	#[strum(serialize = "accelerator label")]
	AcceleratorLabel,
```

DataTriny:
> You mentioned we get State and Role were obtained from EventBody , I presume you meant the Message's body then

For the `StateChanged` signal a `State` has to fit into the `kind` field, and in the case of the `PropertyChange` signal a `Role` must go into `any_data`. This might get abstracted away on the client side by the event stream, but on the server side I will have to construct these `EventBody` manually once I update AccessKit.

> I meant the scheme with spaces.

So you are talking about the value we are supposed to return from the `GetRoleName` method of the `org.a11y.atspi.Accessible`. We don't implement it on AccessKit and I don't plan to do so as I think it is pointless. I might convince the GNOME team to deprecate it at some point.

TTWNO:
> GetRoleName method of the org.a11y.atspi.Accessible interface. We don't implement it on AccessKit and I don't plan to do so as I think it is pointless. I might convince the GNOME team to deprecate it at some point.

Completely agree. Localization shouldn't be done by sending messages over DBus. `GetRole` is sufficient.

luukvanderduim:
> > You mentioned we get State and Role were obtained from EventBody , I presume you meant the Message's body then
> 
> For the `StateChanged` signal a `State` has to fit into the `kind` field, and in the case of the `PropertyChange` signal a `Role` must go into `any_data`. 

Yes, of course. 
I realized I was mistaken there  some minutes after posting.. 

> This might get abstracted away on the client side by the event stream, but on the server side I will have to construct these `EventBody` manually once I update AccessKit.

I think I understand. 
Ideally you would want a `Serialize` implementation for, let's take `Role`, that could serialize to `EventBody`, `EventBodyQt`, `Role`? 

We cannot have that in one `Serialize` impl. 
We can have the type itself, and two newtypes and on the newtypes you could each implement separate `Serialize` impls.

> 
> > I meant the scheme with spaces.
> 
    > So you are talking about the value we are supposed to return from the `GetRoleName` method of the `org.a11y.atspi.Accessible` interface. We don't implement it on AccessKit and I don't plan to do so as I think it is pointless. I might convince the GNOME team to deprecate it at some point.

I did not know you had the strings there for that purpose. 

I think GNOME people need little convincing as it is currently only said to be used for roles outside of the enumeration and itś removal is eluded to in the specs:

```XML
    <!--
        GetRoleName:

        Gets a UTF-8 string corresponding to the name of the role played by an object.
        This method will return useful values for roles that fall outside the
        enumeration used in the GetRole method. Implementing this method is
        optional, and it may be removed in a future version of the API.
        Libatspi will only call id in the event of an unknown role.
    -->
```

TTWNO:
Nice find, Luuk! Good to know.

DataTriny:
Sorry @luukvanderduim I missed your last comment. I think we are both on the same track now! It's not clear what's the solution for me either, maybe we should open a new issue to explore how to improve the event types. But I really think we can go quite far by making full use of serde, without introducing strum.

TTWNO:
You may be right, @DataTriny , but I think having an optional dependency makes sense here. `FromRepr`, and `IntoStaticStr` seem to be really useful abstractions. `Display` is whatever, since that can be done with a `macro_rules!`.

TTWNO:
After making `strum` optional, and gating all uses of it behind the feature "strum", it seems that Strum has become a bit more fundamental on this branch than before. It causes a cascading set of conversion errors all the way up, and so this branch does indeed make `strum` required.

DataTriny:
@TTWNO I'm definitely not advocating for ditching all the work already done by @luukvanderduim with this PR. Maybe strum will be necessary to fully support features such as fine-grained event filtering. But I think we need to first modify our types to make full use of serde, then we can revisit this PR and possibly add strum as an optional dependency if it is only useful to AT-SPI clients for instance. strum might be just one dependency, but it adds more procedural macros, which hurt compile times. Since we already have quite a few in the Rust accessibility stack, I think we should try to avoid it where we can.

TTWNO:
Yeah, I shouldn't merge with `main` but GH interface was complaining.

TTWNO:
@DataTriny 

Curious if this would work for you? We've fully feature-gated `strum`, and still provided the vast majority of the implementations despite that. The only implementations that are not available are `EventTypeProperties` for some specific events. Assuming you don't use this (except for in tests, where I assume you are not as worried about compile times and deps) this should work out well.

Let me know.

DataTriny:
@TTWNO Thanks for all the efforts you put into this! Unfortunately this PR seem to contain way too many unrelated changes for me to review right now, but I checked that indeed the common crate can compile without the strum feature. However you have tied it to the features for async runtime selection, which mean that AccessKit would still pull in strum as we use a few proxies from atspi-proxies. Is this a requirement, or could an async executor be enabled without strum?

TTWNO:
@DataTriny right now, any use of `zbus` brings in the `TryFrom/TryInto<Message>` impls for each event. These require strum.

Does accesskit use these implementations? If not, we can add a *separate* feature flag for the message conversions separate from zbus support from proxies.

DataTriny:
No we currently don't use these, although I was eventually planning on doing so. If feature-gating these convertions is the only way forward, I'm fine with what you propose.

TTWNO:
If you intend on using the `TryFrom/TryInto` implementations eventually, and they require `strum` does that mean that `acceskit` is ok to depend on strum?

If not, does that mean that you will write your own `Message` (de)serializers?

DataTriny:
For now we create messages to emit signals by hand. Since the code is already there I will probably leave it as is if relying on atspi would bring strum.

luukvanderduim:
Could we close this? I mean - without looking - I don't believe we're using strum or need it?

DataTriny:
So there is now no easy way to emit a state change event right? With the old `EventBody` type, the `kind` field was generic over `Serialize` so I could put `State` enum variants in there.

PR #174, opened 2024-04-14
Use `serde` derive macros on `Interface`

DataTriny:
Relates to #157 

There are more complex cases where avoiding the use of the `strum` crate will require more design changes, so I'm addressing this one first.

### DIFF
--- a/atspi-common/src/interface.rs
+++ b/atspi-common/src/interface.rs
@@ -5,156 +5,78 @@
 
 use enumflags2::{bitflags, BitFlag, BitFlags};
 use serde::{
-	de::{self, Deserialize, Deserializer, Visitor},
-	ser::{Serialize, Serializer},
+	de::{self, Deserializer, Visitor},
+	ser::{self, Serializer},
+	Deserialize, Serialize,
 };
 use std::fmt;
 use zvariant::{Signature, Type};
 
-const ACCESSIBLE_INTERFACE_NAME: &str = "org.a11y.atspi.Accessible";
-const ACTION_INTERFACE_NAME: &str = "org.a11y.atspi.Action";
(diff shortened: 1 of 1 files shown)


DataTriny:
The `codecov` CI job is failing, not because of a decrease in coverage though.

FWIW I have removed some tests related to the manual serialization/deserialization of the `Interface` type so I expect the coverage to decrease a bit.

TTWNO:
> The codecov CI job is failing, not because of a decrease in coverage though.

No problem. It's always a bit tempermental.

atspi-common/src/interface.rs
```diff
@@ -269,64 +191,6 @@ impl std::ops::BitOr for InterfaceSet {
...
-			VALUE_INTERFACE_NAME => Ok(Interface::Value),
-			REGISTRY_INTERFACE_NAME => Ok(Interface::Registry),
-			DEVICE_EVENT_CONTROLLER_INTERFACE_NAME => Ok(Interface::DeviceEventController),
-			DEVICE_EVENT_LISTENER_INTERFACE_NAME => Ok(Interface::DeviceEventListener),
-			_ => Err("No interface found for conversion."),
-		}
-	}
-}
-
-impl std::fmt::Display for Interface {
```
TTWNO, line 302: Why have you removed the `Display` implementation? Does `Serialize` auto-implement this?

DataTriny: I removed the constants so as to not have duplicate names, so we'd have to write the string literals inside the `Display` impl in order to keep this. But what's the added value of having this? It was only used in tests that I removed, I think the `Debug` impl is enough for users.

TTWNO:
This removes a ton of manual code. I'm merging.

PR #175, opened 2024-04-14
Remove some unused dependencies of `atspi-proxies`

DataTriny:
Not sure how we ended up keeping this around.


TTWNO:
Since it was optional, it wasn't even caught by the CI.

PR #177, opened 2024-04-18
Make ObjectRef more explicit in type

TTWNO:
Make ObjectRef more explicit in that the "name" refers to a `zbus_names::BusName`, and not just any String.

### DIFF
--- a/atspi-common/src/cache.rs
+++ b/atspi-common/src/cache.rs
@@ -4,7 +4,8 @@
 use crate::{InterfaceSet, ObjectRef, Role, StateSet};
 use serde::{Deserialize, Serialize};
 use zbus_lockstep_macros::validate;
-use zvariant::Type;
+use zbus_names::BusName;
+use zvariant::{ObjectPath, Type};
 
 /// The item type provided by `Cache:Add` signals
 #[allow(clippy::module_name_repetitions)]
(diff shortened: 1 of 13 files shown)


atspi-common/src/cache.rs
```diff
@@ -37,15 +37,15 @@ impl Default for CacheItem {
 	fn default() -> Self {
 		Self {
 			object: ObjectRef {
-				name: ":0.0".into(),
+				name: ":0.0".try_into().unwrap(),
```
DataTriny, line 40: I think it's acceptable to use `from_static_str` in cases like this. There's no need to bring error handling when the value is known at compile time.

DataTriny (changes requested):

DataTriny (approved):
Looks good to me! The amount of repetitive code here is quite annoying though...

TTWNO:
> Looks good to me! The amount of repetitive code here is quite annoying though...

Addressed by #179 , where `GenericEvent` is split up into smaller traits, which are much, much easier to implement via `macro_rules!` and blanket impls.

I've already requested a review from Luuk, but I'd appreciate a look from yourself as well.

DataTriny:
Will try to do it tomorrow.

PR #179, opened 2024-04-18
new generic event trait

TTWNO:
Depends on #177 

Still TODO before ready:

- [x] split additional traits out for name+path
- [x] all but one related trait should be object-safe
	- [x] and enforce with `static_assertions` (no new deps, we already depend on this)
- [x] Implement new interfaces on wrapper types.
- [x] test that event introspection/transparency works correctly.


TTWNO:
I just implemented these manually. `enum_dispatch` is annoying and require _it_ to implement `From` for you inner types.

TTWNO:
For those reviewing:

I'm particularly interested in:

1. Did I choose the right names.
2. Are the traits split up in a fashion that makes sense, intuitively.

TTWNO:
Forgot that Github's "resolve conflicts" feature merges from `main` instead of rebasing; solved by a force-pushing.

PR #182, opened 2024-04-28
Obtain an `AccessibleProxy` from any `ObjectRef` with `ObjectRefExt`

luukvanderduim:
It is assumed that all objects in any server's tree implements the `Accessible` interface.
Therefore it would be very convenient to quickly obtain a `AccessibleProxy` on an object. 

Bringing `ObjectRefExt` into scope allows users to call `as_accessible_proxy` on any `ObjectRef` to obtain an `AccessibleProxy` from the `ObjectRef`.

### DIFF
--- a/atspi-proxies/src/accessible.rs
+++ b/atspi-proxies/src/accessible.rs
@@ -249,8 +249,7 @@ impl TryFrom<&AccessibleProxy<'_>> for ObjectRef {
 	}
 }
 
-#[allow(dead_code)]
-trait ObjectRefExt {
+pub trait ObjectRefExt {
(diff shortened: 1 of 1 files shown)


luukvanderduim:
We cannot do the same with `TryFrom<ObjectRef> for AccessibleProxy` because the proxy builder needs a `zbus::Connection` as argument.

This is probably the best we can do.

TTWNO:
This is a great usability improvement for users of atspi, and only adds a few lines of code.

Since this is somewhat non-trivial to test, I'm going to ignore codecov.

PR #183, opened 2024-05-03
`from_message_parts` into `from_message`

luukvanderduim:
Builds on 'new-generic-event-trait'

`from_message` takes just one argument, `&zbus::Message` and converts into the `BusProperties` implementing type.

This avoids crreating a body in the cases we do not need one.

This also introduces a `MessageExt` trait that provides `matches_event` This allows users who have / prefer a `MessageStream` over the `event_stream` to do something like this:

```Rust
	if msg.matches_event::<AbsEvent>()? {
		AbsEvent::from_message()?
	} else {
		Err(AtspiError::EventMisMatch)
	};
```

or simply use:

```Rust
	AbsEvent::try_from(msg)?
```

Which now uses `matches_event`.

Lastly, this has a few little optimizations that avoid clones in a conversion.

### DIFF
--- a/atspi-common/src/error.rs
+++ b/atspi-common/src/error.rs
@@ -30,6 +30,9 @@ pub enum AtspiError {
 	/// No Signature.
 	MissingSignature,
 
+	/// When matching on a signal with an event, but the event does not match the signal.
+	EventMismatch,
+
 	/// When matching on an unknown role
 	UnknownRole(u32),
 
@@ -86,6 +89,7 @@ impl std::fmt::Display for AtspiError {
 			Self::MissingInterface => f.write_str("Missing interface."),
 			Self::MissingMember => f.write_str("Missing member."),
 			Self::MissingSignature => f.write_str("Missing signature."),
+			Self::EventMismatch => f.write_str("Event <--> Message mismatch"),
 			Self::UnknownRole(e) => f.write_str(&format!("atspi: Unknown role: {e}")),
 			Self::UnknownSignal => f.write_str("atspi: Unknown signal"),
 			Self::CacheVariantMismatch => f.write_str("atspi: Cache variant mismatch"),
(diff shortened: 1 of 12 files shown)


TTWNO:
Does it make sense to abstract away the nature of the mismatch? In my mind, it'd be much more useful to have specific descriptions of the mismatch, instead of having library users needing to add a bunch of debugging to figure it out.

Just my thoughts. Let me know.

TTWNO:
Otherwise, this is a welcome change. I'm glad that a rename and redesign was considered.

luukvanderduim:
> Does it make sense to abstract away the nature of the mismatch? In my mind, it'd be much more useful to have specific descriptions of the mismatch, instead of having library users needing to add a bunch of debugging to figure it out.
> 
> Just my thoughts. Let me know.

I lean towards that a mismatch is not an error at all. If it matches 'false', nothing 'bad' happened, the outcome is just not what you want. 
The user registers for particular event(s) and deliberately `msg.matches_event::<AttributesChangedEvent>()`, right?
Maybe you want to match against a different type or use a catch-all like `msh.matches_event::<ObjectEvents>()`.

TTWNO:
Hmmmm.... It looks like this breaks my goal of keeping common WASM compatible. I'm still thinking about this to see if it's strictly necessary, and if it is, how can we still integrate this.

I'm in favour of fewer lines of code, almost always.

luukvanderduim:
> Hmmmm.... It looks like this breaks my goal of keeping common WASM compatible.

Constructing errors breaks WASM? What rule should I not have broken?

 I'm still thinking about this to see if it's strictly necessary, and if it is, how can we still integrate this.

What do 'it' and 'this' refer to? Greetings from the disambiguation dept! 

> 
> I'm in favour of fewer lines of boilerplate code :)
For the user?

TTWNO:
The work is mostly redundant due to #206 

Please let me know if I've made a mistake.

PR #184, opened 2024-05-15
Maintanance fixes

luukvanderduim:
Review with rigor, handle with care, scrutinize like it's 1999!

See individual commits for details.

- [Rename 'accessible' --> 'obj'](https://github.com/odilia-app/atspi/commit/02ceec211b48f53c341cee20ff0a765da0467001) 
- [Fix missing feature in atspi crate](https://github.com/odilia-app/atspi/commit/b5e8cc58bf2d1ca771548967fae8cd2f6cf3cd35)
- [Update dependencies](https://github.com/odilia-app/atspi/commit/6f24fbc65f4dc9020407e6ee4f07954899145a36)
- [Deprecation of "default_features"](https://github.com/odilia-app/atspi/commit/f1696afcb01e94c6036953f2ec43b782e7aecc46)
- [Fix search-and-replace mishap](https://github.com/odilia-app/atspi/commit/35d1bf70414386a103379b751e439c008bca1972)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,26 +1,15 @@
 [workspace]
 resolver = "2"
-members = [
-  "atspi",
-	"atspi-proxies",
-	"atspi-common",
-  "atspi-connection",
-]
+members = ["atspi", "atspi-proxies", "atspi-common", "atspi-connection"]
 
 
 [workspace.package]
 description = "AT-SPI2 protocol implementation in Rust"
-authors = [
-  "Michael Connor Buchan <mikey@blindcomputing.org>", 
-  "Tait Hoyem <tait@tait.tech>", 
-  "Alberto Tirla <albertotirla@gmail.com>", 
-  "DataTriny <datatriny@gmail.com>", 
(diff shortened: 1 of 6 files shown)


TTWNO:
Thank you again Luuk for catching these important issues! I really would have never noticed, because I know the history and would have brushed it off.

I'm curious why you chose to upgrade zbus and zvarinat though? Generally, we want to use the oldest compatible version of dependencies. It would be a shame if we caused a version mismatch downstream just by being too aggressive about upstream upgrades.

luukvanderduim:
[Z-crates Changelogs](https://github.com/dbus2/zbus/releases)

zbus 4.2.1 has a number of "Critical fixes", especially the one where properties are invalidated after querying(?) seems problematic. 
Also "Hack to make connection handshake work inside Flatpak. xdg-dbus-proxy" might be of importance for us.
zbus 4.2.0 dropped two dependencies.

I chose to also upgrade zvariant to keep the z-crates underlying dependencies in sync.

But if you want to investigate this a bit further, I am fine if you drop this commit and just take the others.

PR #186, opened 2024-05-22
Stricten type of ObjectRef::name

TTWNO:
- Accoring to DBus specification, there is always a `UniqueName` style bus name.[1]

[1]: https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus


TTWNO:
> What's the purpose of the file atspi-common/p1.txt you've included as part of this PR?

Whoopsie; was some old testing data that didn't get cleared.

Issue #173, opened 2024-04-13
Add example for getting the accessibility tree


C-Loftus:
I believe that getting the accessibility tree is very important for many accessibility applications and would be helpful to show as an example. For instance, for voice based interaction there are many times where we want to annotate all the a11y elements with a label. (i.e. labeling it with a number so we can just say the number to click on it)

The only other atspi example repo on Github shows examples on dumping the tree so I think there is precedent that this is a useful example to show
 https://github.com/infapi00/at-spi2-examples

TTWNO:
Fixed by #188

PR #188, opened 2024-05-26
Add 'bus-tree' example

luukvanderduim:
Addresses issue #173 

- Adds the 'bus-tree' example 

~~Also try out `static_assertions::assert_cfg!` to inform users of the appropriate features to use.
This will make rust-analyzer noisy if you don't have these set in your editor.~~

- Add `[[example]]` section in Cargo.toml, which allows us to point out, using the `required-features` list which features a user needs.
 
~~This tries to add `.vscode/settings.json` to mitigate that but if this is deemed too opinionated, just remove it.~~ 
removed, no longer needed.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,0 @@
-{
-    "rust-analyzer.cargo.features": [
-        "zbus",
-        "proxies-tokio",
-        "tokio"
-    ]
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+    "rust-analyzer.cargo.features": [
+        "zbus",
+        "proxies-tokio",
+        "tokio"
+    ]
+}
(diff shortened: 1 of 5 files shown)


TTWNO:
@luukvanderduim is it possible, knowing what we know able Libreoffice Calc (2^31 children-cells), could you please add functionality for modelling "countable but almost infinite children"?

luukvanderduim:
> @luukvanderduim is it possible, knowing what we know able Libreoffice Calc (2^31 children-cells), could you please add functionality for modelling "countable but almost infinite children"?

Yeah.. it comes at the cost of adding a call per child to get the number of children.

A reasonable number:
 - 65536 children (2^16) Children, which is max_descendants in the `Object` interface.
 
 A legal threshold:
 - 128*1024*1024 (Max Message size) / 48 (Size of an ObjectRef) = 2_708_821 children.

Or without the extra call. This will cover all stalling responses not just for (in practical terms) infinite children.

 -  A timeout (eg. 1 sec.)

TTWNO:
> 65536 children (2^16) Children, which is max_descendants in the Object interface.

This seems good to me.

luukvanderduim:
Last commit confirmed to make the example resilient against an 'infinite' number of children.

PR #189, opened 2024-06-09
Fix QT problems (again)

TTWNO:
- Add new `From<EventBodyOwned> for EventBodyQT`
- Use booleans in call to macros:
	- `event_test_case!($ty) -> event_test_case!($ty, true)`
	- `impl_from_dbus_message!($ty) -> impl_from_dbus_message!($ty,
	  true)`
	- `false` should be used as the second argument if the `<$ty as
	  BusProperties>::Body` is not `EventBodyOwned`.
- This only marginally changes the macro expansion of
  `impl_from_dbus_message`, and so we should consider only changing the
  one line instead of having two entirely separate impl blocks.

cc @luukvanderduim

### DIFF
--- a/atspi-common/src/events/mod.rs
+++ b/atspi-common/src/events/mod.rs
@@ -428,8 +428,8 @@ impl_try_from_event_for_user_facing_type!(
 	CacheEvents::LegacyAdd,
 	Event::Cache
 );
-event_test_cases!(LegacyAddAccessibleEvent, false);
-impl_from_dbus_message!(LegacyAddAccessibleEvent, false);
+event_test_cases!(LegacyAddAccessibleEvent, Explicit);
+impl_from_dbus_message!(LegacyAddAccessibleEvent, Explicit);
 impl_event_properties!(LegacyAddAccessibleEvent);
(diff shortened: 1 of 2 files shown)


atspi-common/src/events/mod.rs
```diff
@@ -417,8 +428,8 @@ impl_try_from_event_for_user_facing_type!(
 	CacheEvents::LegacyAdd,
 	Event::Cache
 );
-event_test_cases!(LegacyAddAccessibleEvent);
-impl_from_dbus_message!(LegacyAddAccessibleEvent);
+event_test_cases!(LegacyAddAccessibleEvent, false);
```
luukvanderduim, line 431: We spoke about replacing the bool by a more descriptive enum.

luukvanderduim (commented):
Lets change the bool into an enum and merge this quickly.

In a practical sense I think the added test is perhaps the most important guard to prevent repetition, thanks!

atspi-common/src/events/mod.rs
TTWNO: Now using `Auto/Explicit`

PR #192, opened 2024-06-16
Change get_mdiz_order method call name with attribute; fix #190

### LINKED ISSUE #190: `Component::get_mdizorder()` gets called as `GetMdizorder` instead of `GetMDIZOrder`
from atspi-proxies/src/component.rs

```Rust
	/// GetMDIZOrder method
	fn get_mdizorder(&self) -> zbus::Result<i16>;
```

TTWNO:


TTWNO:
@luukvanderduim Can you please confirm that this works?

luukvanderduim:
> @luukvanderduim Can you please confirm that this works?

This adaptation of fama11y-tree tries to call `get_mdiz_order` on all objects that support `Component`.
Looks like the call works, though we only gain a 'now working call' as no single object seems to report anything over 0.

https://github.com/luukvanderduim/fama11y-tree/tree/rename-mdiz-layer-method-test

PR #193, opened 2024-06-17
Add option to initialize atspi connection with custom queue size

TTWNO:
I've run into issues with Odilia whereby the lack of a bigger queue causes a deadlock (this is because the queue gets full, which blocks any task attempting to query).

Adding a larger queue size fixes the problem entirely, since we do poll fairly regularily, but we need some leniency during event storms.


luukvanderduim:
It seems the longer queue masked the issue of not polling continuously?

TTWNO:
Indeed...

PR #194, opened 2024-06-20
Split ChildrenChanged/TextChanged events into distinct Inserted/Deleted events

TTWNO:
I think this works better than adding an enum like "Inserted/Deleted"; they really are separate events. This is obvious if you look at the name of the other events like "ColumnInserted"/"ColumnDeleted", etc. The only thing I'm curious about is how to test this; our automated tests do not allow modification of the body. We can add manual test as well: of course there is nothing wrong with that!

Commits:

- **Split TextChangedEvent into TextInserted/TextDeleted**
- **Split ChildrenChangedEvent into ChildrenInserted/ChildrenDeleted**

TODOs:

- [ ] Test the "kind" field fails when it should, succeedes when it should

### DIFF
--- a/atspi-common/src/events/object.rs
+++ b/atspi-common/src/events/object.rs
@@ -18,8 +18,10 @@ pub enum ObjectEvents {
 	LinkSelected(LinkSelectedEvent),
 	/// See: [`StateChangedEvent`].
 	StateChanged(StateChangedEvent),
-	/// See: [`ChildrenChangedEvent`].
-	ChildrenChanged(ChildrenChangedEvent),
+	/// See: [`ChildrenInsertedEvent`].
+	ChildrenInserted(ChildrenInsertedEvent),
+	/// See: [`ChildrenRemovedEvent`].
+	ChildrenRemoved(ChildrenRemovedEvent),
 	/// See: [`VisibleDataChangedEvent`].
 	VisibleDataChanged(VisibleDataChangedEvent),
 	/// See: [`SelectionChangedEvent`].
@@ -65,7 +67,8 @@ impl EventTypeProperties for ObjectEvents {
 			Self::BoundsChanged(inner) => inner.member(),
 			Self::LinkSelected(inner) => inner.member(),
 			Self::StateChanged(inner) => inner.member(),
-			Self::ChildrenChanged(inner) => inner.member(),
+			Self::ChildrenInserted(inner) => inner.member(),
+			Self::ChildrenRemoved(inner) => inner.member(),
 			Self::VisibleDataChanged(inner) => inner.member(),
 			Self::SelectionChanged(inner) => inner.member(),
 			Self::ModelChanged(inner) => inner.member(),
@@ -92,7 +95,8 @@ impl EventTypeProperties for ObjectEvents {
 			Self::BoundsChanged(inner) => inner.interface(),
 			Self::LinkSelected(inner) => inner.interface(),
 			Self::StateChanged(inner) => inner.interface(),
-			Self::ChildrenChanged(inner) => inner.interface(),
+			Self::ChildrenInserted(inner) => inner.interface(),
+			Self::ChildrenRemoved(inner) => inner.interface(),
 			Self::VisibleDataChanged(inner) => inner.interface(),
 			Self::SelectionChanged(inner) => inner.interface(),
 			Self::ModelChanged(inner) => inner.interface(),
@@ -119,7 +123,8 @@ impl EventTypeProperties for ObjectEvents {
 			Self::BoundsChanged(inner) => inner.match_rule(),
 			Self::LinkSelected(inner) => inner.match_rule(),
 			Self::StateChanged(inner) => inner.match_rule(),
-			Self::ChildrenChanged(inner) => inner.match_rule(),
+			Self::ChildrenInserted(inner) => inner.match_rule(),
+			Self::ChildrenRemoved(inner) => inner.match_rule(),
 			Self::VisibleDataChanged(inner) => inner.match_rule(),
 			Self::SelectionChanged(inner) => inner.match_rule(),
 			Self::ModelChanged(inner) => inner.match_rule(),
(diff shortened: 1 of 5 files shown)


TTWNO:
Well, scratch that! Looks like the tests caught an error! I had the kind "delete | delete/system" deserailize into the ChildrenInserted type, which indeed is a mistake.

I'm going to `s/Removed/Deleted/g` since that's the name of the discriminant in AT-SPI.

luukvanderduim:
Good you managed to catch and mash this 'deleted' -> removed' bug!
 
From the old [event table @ Linux Foundatioon](https://accessibility.linuxfoundation.org/a11yspecs/atspi/adoc/atspi-events.html):

'kind' major:minor discriminants of events that communicate some change:
| Event major | minor |
| --- | --- |
| text-changed | inserted / removed |
| text-selection-changed | - |
| text-attributes-changed | - |
| property-change| accessible-parent, accessible-name, accessible-description, accessible-value, accessible-role, accessible-table-caption, accessible-table-column-description, accessible-table-column-header, accessible-table-row-description, accessible-table-row-header, accessible-table-summary, |
| state-changed | 'enabled' bool | 
| children-changed | add, remove |
| visible-data-changed | - |
| bounds-changed | - |
| selection-changed | - |
| model-changed | - |
| active-descendants-changed | - |
| selection-changed | - |
| content-changed | - |
| attributes-changed (doc) | - |
| line-changed | - | 
| columncount-changed | - |
| linecount-changed | - |
| application-changed | - |
| charwidth-changed | - |


This does break the 1:1 relationship between signals' member and user facing event types doesn't it?

One could argue that all events with minor discriminants warrant separate events..
How about `StateChanged` with it's 'enabled' field? Why is that different? 
Or `Button` with 'p1', 'r1', 'p2', etc? 
Is there perhaps a particular problem you found that is fixed by separating `ChildrenChanged` and `TextChanged` ?

atspi-common/src/events/object.rs
```diff
@@ -1004,7 +1059,18 @@ impl TryFrom<&zbus::Message> for ObjectEvents {
 			"BoundsChanged" => Ok(ObjectEvents::BoundsChanged(ev.try_into()?)),
 			"LinkSelected" => Ok(ObjectEvents::LinkSelected(ev.try_into()?)),
 			"StateChanged" => Ok(ObjectEvents::StateChanged(ev.try_into()?)),
-			"ChildrenChanged" => Ok(ObjectEvents::ChildrenChanged(ev.try_into()?)),
+			"ChildrenChanged" => {
+				let body: EventBodyOwned = ev.try_into()?;
+				match body.kind.as_str() {
```
luukvanderduim, line 1064: Are these discriminants correct?
In the old [LF event table](https://accessibility.linuxfoundation.org/a11yspecs/atspi/adoc/atspi-events.html) it says these minor strings were 'add' and 'remove' did that change?

atspi-common/src/events/object.rs
```diff
@@ -1019,7 +1085,14 @@ impl TryFrom<&zbus::Message> for ObjectEvents {
 			"ColumnDeleted" => Ok(ObjectEvents::ColumnDeleted(ev.try_into()?)),
 			"TextBoundsChanged" => Ok(ObjectEvents::TextBoundsChanged(ev.try_into()?)),
 			"TextSelectionChanged" => Ok(ObjectEvents::TextSelectionChanged(ev.try_into()?)),
-			"TextChanged" => Ok(ObjectEvents::TextChanged(ev.try_into()?)),
+			"TextChanged" => {
+				let body: EventBodyOwned = ev.try_into()?;
+				match body.kind.as_str() {
```
luukvanderduim, line 1090: As above, but for TextChanged, the discriminants may be 'inserted' and 'removed'?

atspi-common/src/events/object.rs
```diff
@@ -1138,25 +1211,59 @@ impl From<StateChangedEvent> for EventBodyOwned {
...
+event_test_cases!(ChildrenDeletedEvent);
+impl_to_dbus_message!(ChildrenDeletedEvent);
+impl_from_dbus_message!(ChildrenDeletedEvent);
+impl_event_properties!(ChildrenDeletedEvent);
+
+impl From<ChildrenInsertedEvent> for EventBodyOwned {
+	fn from(event: ChildrenInsertedEvent) -> Self {
+		EventBodyOwned {
+			properties: std::collections::HashMap::new(),
+			kind: "insert".to_string(),
```
luukvanderduim, line 1249: Should this not be "add"?

atspi-common/src/events/object.rs
```diff
@@ -1138,25 +1211,59 @@ impl From<StateChangedEvent> for EventBodyOwned {
...
+		}
+	}
+}
+
+impl From<ChildrenDeletedEvent> for EventBodyOwned {
+	fn from(event: ChildrenDeletedEvent) -> Self {
 		EventBodyOwned {
 			properties: std::collections::HashMap::new(),
-			kind: event.operation,
+			kind: "delete".to_string(),
```
luukvanderduim, line 1266: Should this not be "remove"?

I may be misunderstanding things.

luukvanderduim (commented):
Why are the discriminants correct?

atspi-common/src/events/object.rs
luukvanderduim: Okay, I just tried a bit of monitoring on the bus and found that "insert" and "delete" seem right for 'kind'.

The <major>:<minor> values only matter for the match-rule.. So we could construct unique match rules for these.

atspi-common/src/events/object.rs
luukvanderduim: With busmonitoring I saw ChildrenChanged: "add/system", "remove/system" and "add", "remove"
So this should probably be "add" or "add/system"

atspi-common/src/events/object.rs
TTWNO: You are right.

atspi-common/src/events/object.rs
TTWNO: Yes.

PR #196, opened 2024-06-22
Fix our `Collection` story

luukvanderduim:
ToDo:

- [ ] document collection module level
- [ ] document collection methods
- [ ] document collection method arguments.

The above may take some digging in remote sources or asking around.
Note that 'collection.xml' isn't that helpful here.

### DIFF
--- a/atspi-common/src/object_match.rs
+++ b/atspi-common/src/object_match.rs
@@ -1,0 +1,236 @@
+use std::{collections::HashMap, marker::PhantomData};
+
+use serde::{Deserialize, Serialize};
+use zvariant::{Signature, Type};
+
+use crate::{Interface, InterfaceSet, Role, StateSet};
+
+/// Defines how an object-tree is to be traversed.
+/// Used in `CollectionProxy`.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize, Type)]
+#[repr(u32)]
+pub enum TreeTraversalType {
+	/// When traversing the tree, restrict to children of the current object.
+	RestrictChildren,
+
+	/// When traversing the tree, restrict to siblings of the current object.
+	RestrictSibling,
+
+	/// Traverse the tree in order of appearance.
+	#[default]
+	Inorder,
+}
+
+/// Definition of match rules for accessible objects.
+/// Rule(s) against which we can match an  object or a collection thereof.
+///
+/// # Examples
+///  ```Rust
+///     let builder = MatchRule::builder();
+/// ```
+///
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
+pub struct ObjectMatchRule {
+	pub states: StateSet,
+	pub states_mt: MatchType,
+	pub attr: HashMap<String, String>,
+	pub attr_mt: MatchType,
+	pub roles: Vec<Role>,
+	pub roles_mt: MatchType,
+	pub ifaces: InterfaceSet,
+	pub ifaces_mt: MatchType,
+	pub invert: bool,
+	// Private phantom, gets compiled away.
+	// Here to ensure the builder is the only route to obtain a `MatchRule`
+	#[serde(skip)]
+	phantom: std::marker::PhantomData<()>,
+}
+
+// !!! WARNING !!! :
+//
+// State and Role are defined as u32 in Accessible.xml but as i32 in Collection.xml
+//
+// The signature on StateSet is defined inconsistently in the XMLs
+// Accessible.xml: GetState type="au"
+// Collection.xml: GetMatches argument name="rule" type="(aiia{ss}iaiiasib)"
+// The latter starts with ai, which is a signature for an array of signed ints i32.
+//
+// https://gitlab.gnome.org/federico/at-spi2-core/-/commit/4885efedeef71e0df8210622771a0b1bb10e194d
+impl Type for ObjectMatchRule {
+	fn signature() -> Signature<'static> {
+		Signature::from_str_unchecked("(aiia{ss}iaiiasib)")
+	}
+}
+
+impl ObjectMatchRule {
+	/// Create a new `MatchRuleBuilder`
+	#[must_use]
+	pub fn builder() -> ObjectMatchRuleBuilder {
+		ObjectMatchRuleBuilder::default()
+	}
+}
+
+/// The 'builder' type for `MatchRule`.  
+/// Use its methods to set match criteria.
+#[derive(Debug, Clone, Default)]
+pub struct ObjectMatchRuleBuilder {
+	states: StateSet,
+	states_mt: MatchType,
+	attr: HashMap<String, String>,
+	attr_mt: MatchType,
+	roles: Vec<Role>,
+	roles_mt: MatchType,
+	ifaces: InterfaceSet,
+	ifaces_mt: MatchType,
+	invert: bool,
+}
+
+impl ObjectMatchRuleBuilder {
+	/// Insert a `StateSet` to the builder
+	#[must_use]
+	pub fn states(mut self, state_set: StateSet, mt: MatchType) -> Self {
+		self.states = state_set;
+		self.states_mt = mt;
+		self
+	}
+
+	/// Insert a map of attributes
+	#[must_use]
+	pub fn attributes(mut self, attributes: HashMap<String, String>, mt: MatchType) -> Self {
+		self.attr = attributes;
+		self.attr_mt = mt;
+		self
+	}
+
+	/// Insert a slice of `Role`s
+	#[must_use]
+	pub fn roles(mut self, roles: &[Role], mt: MatchType) -> Self {
+		self.roles = roles.into();
+		self.roles_mt = mt;
+		self
+	}
+
+	/// Insert a slice of `Interface`s
+	#[must_use]
+	pub fn interfaces(mut self, interfaces: &[Interface], mt: MatchType) -> Self {
+		self.ifaces = interfaces.iter().copied().collect::<InterfaceSet>();
+		self.ifaces_mt = mt;
+		self
+	}
+
+	/// Sets the inversion of the `MatchRule`, defaults to `false`, no inversion.
+	#[must_use]
+	pub fn invert(mut self, invert: bool) -> Self {
+		self.invert = invert;
+		self
+	}
+
+	/// Builds the [`ObjectMatchRule`]
+	///
+	/// [`ObjectMatchRule``]: crate::object_match::ObjectMatchRule
+	#[must_use]
+	pub fn build(self) -> ObjectMatchRule {
+		ObjectMatchRule {
+			states: self.states,
+			states_mt: self.states_mt,
+			attr: self.attr,
+			attr_mt: self.attr_mt,
+			roles: self.roles,
+			roles_mt: self.roles_mt,
+			ifaces: self.ifaces,
+			ifaces_mt: self.ifaces_mt,
+			invert: self.invert,
+			phantom: PhantomData,
+		}
+	}
+}
+
+/// Enumeration used by [`MatchArgs`] to specify how to interpret [`ObjectRef`] objects.
+///
+/// [`ObjectRef`]: crate::object_ref::ObjectRef
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Type, Default)]
+#[repr(i32)]
+pub enum MatchType {
+	#[default]
+	/// Invalidates match criterion.
+	Invalid,
+
+	/// All of the criteria must be met.
+	All,
+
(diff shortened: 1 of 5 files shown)


TTWNO:
Just took a cursory glance and am very excited! I can review this today, but probably still a couple hours out. Only thing I noticed is that there's link as to *why* no more then N objects will be returned from certain methods.

Thank you  Luuk! This looks like quite the upgrade!

luukvanderduim:
> Just took a cursory glance and am very excited! I can review this today, but probably still a couple hours out. Only thing I noticed is that there's link as to _why_ no more then N objects will be returned from certain methods.
> 

You raise a better point than I would like.

In atk-adaptor of at-spi2-core, we find that 
```C
#define MAX_CHILDREN 65536
```
and both implemented sort-orders check for it.

However, atk-adaptor may not be the only implementer on the bus.
And older versions may not even implement that so there is nothing that guarantees this check.

Accesskit does not implement `Collection`.

> Thank you Luuk! This looks like quite the upgrade!

Thanks!
[Joanmarie shared positive `Collection` experiences in the old calc-issue](https://bugs.documentfoundation.org/show_bug.cgi?id=156657), this motivated me to look into what we needed to get going. 
The `ObjectMatchRule` and builder I still had lying around so what was left was integration, cleanups and - in this case very important - documentation. 

What we need is a few experiments to see if things indeed work as we think it does.

atspi-common/src/object_match.rs
```diff
@@ -0,0 +1,281 @@
...
+	#[must_use]
+	pub fn roles(mut self, roles: &[Role], mt: MatchType) -> Self {
+		self.roles = roles.into();
+		self.roles_mt = mt;
+		self
+	}
+
+	/// Insert a slice of `Interface`s
+	#[must_use]
+	pub fn interfaces(mut self, interfaces: &[Interface], mt: MatchType) -> Self {
```
TTWNO, line 115: Possibly consider `interfaces: T where T: IntoIteer<Interface>`? I think we've implemented that on `InterfaceSet`.

atspi-common/src/object_match.rs
```diff
@@ -0,0 +1,281 @@
...
+	}
+}
+
+/// Enumeration used by [`ObjectMatchRule`] to specify how to interpret [`ObjectRef`] objects.
+///
+/// [`ObjectRef`]: crate::object_ref::ObjectRef
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Type, Default)]
+#[repr(i32)]
+pub enum MatchType {
+	#[default]
```
TTWNO, line 154: Unsure if this should be the default. In my mind, a default `ObjectMatch` will match on _all_ objects. Perhaps this is sort of a mistaken idea, but that's where it comes from.

atspi-common/src/object_match.rs
```diff
@@ -0,0 +1,281 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Type)]
+#[repr(u32)]
+/// Enumeration used by interface `CollectionProxy` to specify the way [`ObjectRef`]
+/// objects should be sorted.
+///
+/// # Note
+///
+/// Only `Canonical` and `ReverseCanonical` are currently implemented.
+///
+/// see: [issue 140. "Are all the AtspiCollectionSortOrder types really a thing?"](https://gitlab.gnome.org/GNOME/at-spi2-core/-/issues/140)
```
TTWNO, line 181: Thanks for adding docs to the AT-SPI bridge :)

atspi-common/src/object_match.rs
```diff
@@ -0,0 +1,281 @@
...
+///
+/// [`ObjectRef`]: crate::object_ref::ObjectRef
+pub enum SortOrder {
```
TTWNO, line 184: Another question: what happens if we run code that uses one of these unimplemented sort orders? Does it fail with a DBus message or send zero items? If it fails, I feel like there must be a better way to restrict usage of the variants that we _already_ know will instantly fail: I'm thinking a different-but-similar solution to `#[deprecated]` that will warn automatically when you use a marker variant of an enum.

atspi-common/src/object_match.rs
```diff
@@ -0,0 +1,281 @@
...
+	#[test]
+	fn validate_match_rule_signature() {
+		let signature = method_args_signature!(member: "GetMatchesTo", interface: "org.a11y.atspi.Collection", argument: "rule");
+		assert_eq!(ObjectMatchRule::signature(), signature);
+	}
+
+	#[test]
+	fn validate_match_type_signature() {
+		let rule_signature = method_args_signature!(member: "GetMatchesTo", interface: "org.a11y.atspi.Collection", argument: "rule");
+		let match_type_signature = rule_signature.slice(3..4);
```
TTWNO, line 232: Just add a comment here explaining why you're comparing these random indecies—I know why now, but future me is not as knowledgeable about the internals of `Collection`.

atspi-common/src/object_match.rs
```diff
@@ -0,0 +1,281 @@
...
+		assert_eq!(rule.ifaces, InterfaceSet::default());
+		assert!(!rule.invert);
+	}
+
+	#[test]
+	fn create_object_match_rule() {
+		let rule = ObjectMatchRule::builder()
+			.states(StateSet::new(State::Active), MatchType::All)
+			.attributes(
+				[("name".to_string(), "value".to_string())].iter().cloned().collect(),
```
TTWNO, line 264: The generic over `.attributes` could help here with such a verbose incantation.

atspi-common/src/object_match.rs
```diff
@@ -0,0 +1,281 @@
...
+			)
+			.roles(&[Role::Alert], MatchType::All)
+			.interfaces(&[Interface::Action], MatchType::Any)
+			.invert(true)
+			.build();
+
+		assert_eq!(rule.states, StateSet::new(State::Active));
+		assert_eq!(
+			rule.attr,
+			[("name".to_string(), "value".to_string())].iter().cloned().collect()
```
TTWNO, line 275: Likewise here as above.

atspi-proxies/src/collection.rs
```diff
@@ -1,60 +1,95 @@
...
 trait Collection {
-	/// GetActiveDescendant method
-	fn get_active_descendant(&self) -> zbus::Result<ObjectRef>;
+	// The active descendant of the given object.
+	//
+	// Looks like this is unimplemented.
+	//
+	// See [atspi/collection.c](https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/main/atspi/atspi-collection.c?ref_type=heads#L272)
+	//
+	// fn get_active_descendant(&self) -> zbus::Result<ObjectRef>;
```
TTWNO, line 27: Good find and good call disabling the method.

atspi-proxies/src/collection.rs
```diff
@@ -1,60 +1,95 @@
...
+	///
+	/// # Arguments
+	///
+	/// * `rule` - An [`ObjectMatchRule`] describing the match criteria.
+	/// * `sortby` - A [`SortOrder`] specifying the way the results are to be sorted.
+	/// * `count` - The maximum number of results to return, or 0 for no limit.
+	/// * `traverse` - Not supported.
+	///
+	/// [`ObjectMatchRule`]: atspi_common::object_match::ObjectMatchRule
+	/// [`SortOrder`]: atspi_common::SortOrder
```
TTWNO, line 39: Very thorough and much nicer to read documentation.

atspi-proxies/src/collection.rs
```diff
@@ -1,60 +1,95 @@
...
-	/// GetMatchesFrom method
+	/// Retrieves objects from the collection, after `current_object`, matching a given `rule`.
+	///
+	/// # Arguments
+	///
+	/// * `current_object` - The object at which to start searching.
+	/// * `rule` - An [`ObjectMatchRule`] describing the match criteria.
+	/// * `sortby` - A [`SortOrder`] specifying the way the results are to be sorted.
+	/// * `tree` - A [`TreeTraversalType`] specifying restrictions on the objects to be traversed.
+	/// * `limit_scope` - If `true`, only descendants of `current_object`'s parent will be returned. Otherwise (if `false`), any accessible may be returned if it would preceed `current_object` in a flattened hierarchy.
```
TTWNO, line 56: There's no `limit_scope` value accepted by this method. Why is it in the docs?

atspi-proxies/src/collection.rs
```diff
@@ -1,60 +1,95 @@
...
+	/// * `sortby` - A [`SortOrder`] specifying the way the results are to be sorted.
+	/// * `tree` - A [`TreeTraversalType`] specifying restrictions on the objects to be traversed.
+	/// # Maximum number of objects
+	///
+	/// This method will never return more than 65536 objects.
+	///
+	/// [`ObjectMatchRule`]: atspi_common::object_match::ObjectMatchRule
+	/// [`SortOrder`]: atspi_common::SortOrder
+	/// [`TreeTraversalType`]: atspi_common::TreeTraversalType
+	#[allow(clippy::too_many_arguments)]
```
TTWNO, line 88: There is a `limit_scope` here, so perhaps these functions got their docs reversed?

TTWNO (changes requested):

atspi-common/src/object_match.rs
```diff
@@ -0,0 +1,281 @@
...
+// Collection.xml: GetMatches argument name="rule" type="(aiia{ss}iaiiasib)"
+// The latter starts with ai, which is a signature for an array of signed ints i32.
+//
+// https://gitlab.gnome.org/federico/at-spi2-core/-/commit/4885efedeef71e0df8210622771a0b1bb10e194d
+impl Type for ObjectMatchRule {
+	fn signature() -> Signature<'static> {
+		Signature::from_str_unchecked("(aiia{ss}iaiiasib)")
+	}
+}
+
```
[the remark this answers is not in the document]
luukvanderduim: Its implications are mostly hypothetical and admittedly remote. 

* `StateSet` 

StateSet is a bitfield. As long as each side sticks to bitwise operations only, I think we are all good.

In order for this to go off the rails, imagine the implementer of `Collection` does arithmetic operations on the constituents to retrieve individual states, then the implications depend on what exactly they it is they do, but from our perspective undefined behaviour.

* `Role`
I suspect the number of roles will never exceed 2^31, so no problems.

* Sentinel values
Not problematic in this particular case, because we are  sending a u32, but consider the habit of using -1 to encode something special.and receiving an i32, which you encode as u32.  In two's complement -1 = 11111111 11111111 11111111 11111111
Equals `u32::MAX`, which is many states (in a state set) or likely an illegal or unwanted value for an enum.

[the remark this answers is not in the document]
luukvanderduim: My OCD makes me think protocols should be consistent across the board.

atspi-common/src/object_match.rs
luukvanderduim: Great idea - thanks!

atspi-common/src/object_match.rs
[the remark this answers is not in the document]
TTWNO: > My OCD makes me think protocols should be consistent across the board.

Option types on DBus would help a lot....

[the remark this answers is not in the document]
luukvanderduim: Could you explain how that would help here.

atspi-common/src/object_match.rs
luukvanderduim: Done!

atspi-common/src/object_match.rs
luukvanderduim: Implemented!

atspi-proxies/src/collection.rs
luukvanderduim: Accidentally swapped some around. 
Now fixed!

atspi-proxies/src/collection.rs
luukvanderduim: Yes, they look alike.

atspi-common/src/object_match.rs
```diff
@@ -0,0 +1,281 @@
...
+	pub attr_mt: MatchType,
+	pub roles: Vec<Role>,
+	pub roles_mt: MatchType,
+	pub ifaces: InterfaceSet,
+	pub ifaces_mt: MatchType,
+	pub invert: bool,
+	// Private phantom, gets compiled away.
+	// Here to ensure the builder is the only route to obtain a `MatchRule`
+	#[serde(skip)]
+	phantom: std::marker::PhantomData<()>,
```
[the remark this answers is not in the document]
luukvanderduim: I think I've seen both in the wild? 
Renamed it `_marker`.

TTWNO (approved):

PR #197, opened 2024-06-26
Add Operation Enum

TTWNO:
- **Add new Operation type for add/delete string types**
- **Add KindMatch error variant**
- **Import Operation for crate**
- **Use Operation in TextChanged/ChildrenChangedEvent**

### DIFF
--- a/atspi-common/src/operation.rs
+++ b/atspi-common/src/operation.rs
@@ -1,0 +1,33 @@
+/// An operation can either be [`Self::Insert`] or [`Self::Delete`].
+/// These correspond to methods available on [`Vec`].
+#[derive(Debug, PartialEq, Clone, serde::Serialize, serde::Deserialize, Eq, Hash, Default)]
+pub enum Operation {
+	#[default]
+	#[serde(rename = "add")]
+	#[serde(alias = "add/system")]
+	Insert,
+	#[serde(rename = "delete")]
+	#[serde(alias = "delete/system")]
+	Delete,
+}
+
+impl TryFrom<&str> for Operation {
+	type Error = crate::AtspiError;
+	fn try_from(s: &str) -> Result<Operation, Self::Error> {
+		match s {
+			"add" | "add/system" => Ok(Operation::Insert),
+			"delete" | "delete/system" => Ok(Operation::Delete),
+			_ => Err(crate::AtspiError::KindMatch(format!("\"{s}\" is not a type of Operation"))),
+		}
+	}
+}
+
+impl Into<String> for Operation {
+	fn into(self) -> String {
+		match self {
+			Self::Insert => "add",
+			Self::Delete => "delete",
+		}
+		.to_string()
+	}
+}
--- a/atspi-common/src/operation.rs
+++ b/atspi-common/src/operation.rs
@@ -22,11 +22,11 @@ impl TryFrom<&str> for Operation {
 	}
 }
 
-impl Into<String> for Operation {
-	fn into(self) -> String {
-		match self {
-			Self::Insert => "add",
-			Self::Delete => "delete",
+impl From<Operation> for String {
+	fn from(op: Operation) -> String {
+		match op {
+			Operation::Insert => "add",
+			Operation::Delete => "delete",
 		}
 		.to_string()
 	}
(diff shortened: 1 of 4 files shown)


atspi-common/src/operation.rs
```diff
@@ -0,0 +1,33 @@
+/// An operation can either be [`Self::Insert`] or [`Self::Delete`].
+/// These correspond to methods available on [`Vec`].
+#[derive(Debug, PartialEq, Clone, serde::Serialize, serde::Deserialize, Eq, Hash, Default)]
+pub enum Operation {
+	#[default]
+	#[serde(rename = "add")]
+	#[serde(alias = "add/system")]
```
luukvanderduim, line 7: Serde has so much tricks up its sleeve.. Nice.

atspi-common/src/operation.rs
```diff
@@ -0,0 +1,33 @@
...
+	Insert,
+	#[serde(rename = "delete")]
+	#[serde(alias = "delete/system")]
+	Delete,
+}
+
+impl TryFrom<&str> for Operation {
```
luukvanderduim, line 18: stdlib also offers `FromStr` which, if implemented, allows you to do:

```Rust
        let x = "add";
        let x_fromstr = x.parse::<Operation>().map_err(|_| CreateError)?;
```
It may come down to taste, so I'll leave it up to you.

atspi-common/src/operation.rs
```diff
@@ -0,0 +1,33 @@
...
+	fn try_from(s: &str) -> Result<Operation, Self::Error> {
+		match s {
+			"add" | "add/system" => Ok(Operation::Insert),
+			"delete" | "delete/system" => Ok(Operation::Delete),
+			_ => Err(crate::AtspiError::KindMatch(format!("\"{s}\" is not a type of Operation"))),
+		}
+	}
+}
+
+impl Into<String> for Operation {
```
luukvanderduim: About `Into`: 
It is common practice to not implement `Into` because it comes for free with the `From` implementation.
```Rust
impl From<Operation> for String
```
(In argument position however, you generally do want to use `Into` to catch both `From` and `Into` even if `Into` is implemented explicitly.) 

If possible I'd suggest implementing `Display` for `Operation` instead because it is more versatile. But if you have a specific use-case for the conversion, you could consider implementing `From<Operation> for String` instead?

luukvanderduim (changes requested):
As a rule of thumb, going from types to `String`s, string types or buffers, implement `Display` rather than `From`/`Into`, but it may be you have special considerations?

atspi-common/src/operation.rs
```diff
@@ -10,25 +10,27 @@ pub enum Operation {
...
-			_ => Err(crate::AtspiError::KindMatch(format!("\"{s}\" is not a type of Operation"))),
+			"add" | "add/system" | "insert" | "insert/system" => Ok(Operation::Insert),
+			"delete" | "delete/system" | "remove" | "remove/system" => Ok(Operation::Delete),
+			_ => Err(crate::AtspiError::KindMatch(format!("{s} is not a type of Operation"))),
 		}
 	}
 }
 
 impl From<Operation> for String {
 	fn from(op: Operation) -> String {
```
luukvanderduim, line 30: I left a comment on this implementation on an earlier incarnation of this, which now seems hidden.

As rule of thumb, generally prefer implementing `Display` over `From<Operation> for String` because `Display` is more versatile: [It allows for writing in any buffer - even in no `alloc` situations.  Also, `.to_string` comes for free with `Display` implementations.](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=8153b380f51064a65ef5b43bbcac1f50)

Demonstrated it on the playground.

luukvanderduim (changes requested):
Lets implement `Display` instead!

TTWNO:
Maybe you have to resolve the comments to allow me to run CI again?

luukvanderduim (approved):

luukvanderduim:
> Maybe you have to resolve the comments to allow me to run CI again?

"approve changes" worked. That option was hidden behind three dots.

PR #198, opened 2024-06-26
Enabled Is a Bool

### LINKED ISSUE #45: `.enabled()` property on StateChangedEvent should be a boolean
Whether the StateChangedEvent contains a boolean or not over the DBus wire, it appears that right now it works with an i32 (DBus: "i") return type.

Should this be manually changed in the `identify.rs` file, then add a test for it returning a boolean, or should we leave it up to the library users figure it out? It's not like checking `==1` is super complicated.

TTWNO:
- **Make StateChangedEvent.enabled a bool**
- **Use new bool in tests**

Fix #45

### DIFF
--- a/atspi-common/src/events/object.rs
+++ b/atspi-common/src/events/object.rs
@@ -416,12 +416,33 @@ pub struct StateChangedEvent {
 	pub item: crate::events::ObjectRef,
 	/// The state to be enabled/disabled.
 	pub state: State,
-	/// Enabled or disabled the state.
-	///
-	/// 1 == enabled
-	///
-	/// 0 == disabled
-	pub enabled: i32,
+	/// Whether the state was enabled or disabled.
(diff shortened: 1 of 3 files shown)


atspi-common/src/events/object.rs
```diff
@@ -416,12 +416,33 @@ pub struct StateChangedEvent {
...
 	/// The state to be enabled/disabled.
 	pub state: State,
-	/// Enabled or disabled the state.
-	///
-	/// 1 == enabled
-	///
-	/// 0 == disabled
-	pub enabled: i32,
+	/// Whether the state was enabled or disabled.
+	#[serde(with = "i32_bool_conversion")]
```
luukvanderduim, line 420: Interesting! I had no idea this way of custom deserialization of a field existed.

luukvanderduim (approved):
Looks fine!

PR #202, opened 2024-06-27
Add lazy, cheaply guarded proxy-conversions.

luukvanderduim:
This introduces lazy and cheaply guarded proxy-conversions.

It adds `ObjectProxies` which extends `AccessibleProxy

We have an easy conversion from `ObjectRef` to `AccessibleProxy`

`AccessibleProxy::get_interfaces` gives insight into which interfaces are unokenebted by the object.

`ObjectProxies::proxies` returns `Proxies`, a struct that can contain zero or more proxies to the objects supported interfaces.

The proxies are created only when needed and the check against the `InterfaceSet` is a cheap bitwise operation.

```Rust
// Provided by `ObjectRefExt`
let accessible = obj.as_accessible_proxy();

// Provided by `ObjectProxies` extending `AccessibleProxy`
// Aqcuiring `Proxies` needs to be async to get supported interfaces.
let proxies = accessible.proxies().await?;

{
    let action = proxies.action()?; // init and return &mut `ActionProxy`
    let text = proxies.text()?;
}

let action = proxies.action()?; // already initialized! Return   
```

### DIFF
--- a/atspi-proxies/src/related_proxies.rs
+++ b/atspi-proxies/src/related_proxies.rs
@@ -1,0 +1,317 @@
+use crate::{
+	accessible::AccessibleProxy, action::ActionProxy, application::ApplicationProxy,
+	cache::CacheProxy, collection::CollectionProxy, component::ComponentProxy,
+	document::DocumentProxy, editable_text::EditableTextProxy, hyperlink::HyperlinkProxy,
+	hypertext::HypertextProxy, image::ImageProxy, selection::SelectionProxy, table::TableProxy,
+	table_cell::TableCellProxy, text::TextProxy, value::ValueProxy, AtspiError,
+};
+use atspi_common::{Interface, InterfaceSet, Result};
+
+/// Easily acquire the other interface proxies an object may have.
+///
+/// Equip objects with conversions to proxies of the objects' further implemented interfaces
+/// by extending `AccessibleProxy`.
+///
+/// The `proxies` method returns a `Proxies` object, which contains the related interface proxies.
+///
+/// # Lazy initialization and cheap checks
+///
+/// Proxies are lazily initialized, so they are only created when requested.
+/// Interface availability is checked before creating the proxy and is a cheap bitop.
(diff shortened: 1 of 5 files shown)


TTWNO:
Rebased

atspi-proxies/src/related_proxies.rs
```diff
@@ -0,0 +1,317 @@
...
+/// Equip objects with conversions to proxies of the objects' further implemented interfaces
+/// by extending `AccessibleProxy`.
+///
+/// The `proxies` method returns a `Proxies` object, which contains the related interface proxies.
+///
+/// # Lazy initialization and cheap checks
+///
+/// Proxies are lazily initialized, so they are only created when requested.
+/// Interface availability is checked before creating the proxy and is a cheap bitop.
+pub trait ObjectProxies<'a> {
```
TTWNO, line 21: I'd appreciate something like `ProxyExt` or something like that. Thoughts?

atspi-proxies/src/related_proxies.rs
```diff
@@ -0,0 +1,317 @@
...
+impl<'a> ObjectProxies<'a> for AccessibleProxy<'a> {
+	async fn proxies(&self) -> Result<Proxies<'a>> {
+		let iface_set: InterfaceSet = self.get_interfaces().await?;
+		let proxy = self.inner().clone();
+
+		Ok(Proxies { interfaces: iface_set, proxy, inner: InnerProxies::default() })
+	}
+}
+
+impl<'a> Proxies<'a> {
```
TTWNO, line 62: Implementation is good.

atspi-proxies/src/related_proxies.rs
luukvanderduim: Isn't naming supposed to be the hardest thing in software engineering? 
`ProxyExt` sounds just fine.

TTWNO:
@luukvanderduim are you okay with merging this?

PR #204, opened 2024-06-27
add action struct

TTWNO:
- **Add Action struct**
- **Use Action struct in proxy**


DataTriny:
Perhaps I would move the detailed description of each action components from the interface methods to the new struct fields.

TTWNO:
Can you please explain what you mean? I'm not sure I understand.

PR #205, opened 2024-06-27
Add Display impl for State

TTWNO:
- **Use display instead of Into<String>**
- **Fix some formatting so that syntax highlighting is happy**


luukvanderduim (approved):
Good stuff!

luukvanderduim:
I just saw the clippy run error. 
Easiest fix is to use `State::from_str` instead of `into` there, I think.

DataTriny (approved):
Looks good to me, thanks!

PR #210, opened 2024-11-08
refactor(bench): Dedup bench code and reduce variance between runs

luukvanderduim:
- De duplicate code: 
     Moves common code to module that is shared between tests
- Reduce variance between bench runs: 
    Generates test files with randomly generated `Messages`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 Cargo.lock
+/atspi/benches/resources/*
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 Cargo.lock
-/atspi/benches/resources/*
+/atspi/*.bin
+*.bin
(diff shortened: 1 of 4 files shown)


luukvanderduim:
CI's `cargo machete` complains about `atspi-proxies` depending on `zvariant`. 

Maybe we could try `pub(crate) use zbus/zvariant as zvariant` in the crate and no longer import it ourselves?

TTWNO:
Great PR, Luuk! The best ones are when we remove code! Once #206 lands, I'll rebase this and should be good to go!

TTWNO:
Also your failed checks are because I screwed up the CI, 206 also fixes those problems (I think....)

TTWNO:
Rebased :)

TTWNO (approved):
LGTM!

PR #218, opened 2024-12-17
Event Wrapper Feature Flag

TTWNO:
- [x] Move wrappers from `atspi-common/src/events/*` into `atspi-common/src/events/wrapper/*`
	- [x] Cache
	- [X] Documents
	- [x] Focus
	- [x] Keyboard
	- [x] Mouse
	- [X] Object
	- [x] Terminal
	- [x] Window
- [x] Move all macros/impls related to dispatch and wrapper types into the new modules.
	- [x] Cache
	- [X] Documents
	- [x] Focus
	- [x] Keyboard
	- [x] Mouse
	- [x] Object
	- [x] Terminal
	- [x] Window
- [x] Feature-gate the new wrapper types, their tests, and the macros


TTWNO:
cc @luukvanderduim 

Let me know if this is similar to your idea. Best example is `Document`. Lots of clippy, etc. to fix but just want to make sure I'm on the right path.

luukvanderduim:
> cc @luukvanderduim
> 
> Let me know if this is similar to your idea. Best example is `Document`. Lots of clippy, etc. to fix but just want to make sure I'm on the right path.

Could the following work?

Move these enumeration types into a `event_wrapper.rs` 
Also move all implementations on these enums in this module.
But let's not move end-user facing types in here because those should always be visible.

```
 `enum Event` 
         `enum DocumentEvents`.
	`enum FocusEvents`
	`enum KeyboardEvents`
	`enum MouseEvents`
	`enum ObjectEvents`
	`enum TerminalEvents`
	`enum WindowEvents`
	`enum AvailableEvent`
	`enum CacheEvents`
	`enum EventListenerEvents`
```
also add all implementations on above types, such as:
```Rust
From<&Message> for CacheEvents
```
and (should we have it):
```Rust
From<Event> for DocumentEvents
```
even our big:
```Rust
From<&Message> for Event
```
But, for instance not:
```Rust
From<CacheEvents> for CacheAddEvent
```
People could want to opt out of `Event` and have a `MessageStream` loop and do a 

```Rust
// pseudo rust
while Some(Ok(msg)) = stream.next()? {
   select! {
         CacheAddEvent::try_from(msg)? => got_cache_add(add);
         CaretMovedEvent::try_from(msg)? => caret_moved(moved);
   }
}
```
There is likely a thousand things wrong with above, but the gist is one could write this and not have `Event` around.

TTWNO:
Consider it done. Working on it now.

PR #223, opened 2025-02-25
feat: introduces borrowed bodies

luukvanderduim:
This 'backports' the `event_body` module from our development branch.

- Adds borrowed bodies for qt and common bodies.
- Adds sane defaults and conversions
- Adds `QtProperties` and `Properties` to replace unused types in field 'properties' (HashMaps / ObjectRef types)
- Minimal serialization for `QtProperties` and `Properties`
- Uphold `Type::Signature` requirements for `QtProperties` and `Properties`
- Shrink `EventBody` and `EventBodyQT` from 144 bytes to 96 bytes
- Adds many tests.

In need of eyeballs! ;)

### DIFF
--- a/atspi-common/Cargo.toml
+++ b/atspi-common/Cargo.toml
@@ -1,42 +1,42 @@
 [package]
-name = "atspi-common"
-version = "0.9.0"
-edition = "2021"
-readme = "README.md"
-categories = ["accessibility", "api-bindings"]
-keywords = ["Macros", "Accessibility"]
-repository = "https://github.com/odilia-app/atspi"
-description = "Primitive types used for sending and receiving Linux accessibility events."
-license = "Apache-2.0 OR MIT"
-include = ["src/**/*", "LICENSE-*", "README.md", "xml/*"]
-rust-version.workspace = true
+  categories             = [ "accessibility", "api-bindings" ]
+  description            = "Primitive types used for sending and receiving Linux accessibility events."
+  edition                = "2021"
+  include                = [ "LICENSE-*", "README.md", "src/**/*", "xml/*" ]
+  keywords               = [ "Accessibility", "Macros" ]
+  license                = "Apache-2.0 OR MIT"
+  name                   = "atspi-common"
+  readme                 = "README.md"
+  repository             = "https://github.com/odilia-app/atspi"
+  rust-version.workspace = true
+  version                = "0.9.0"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+  # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["async-std", "wrappers"]
-async-std = ["zbus/async-io"]
-tokio = ["zbus/tokio"]
-wrappers = []
+  async-std = [ "zbus/async-io" ]
+  default   = [ "async-std", "wrappers" ]
+  tokio     = [ "zbus/tokio" ]
+  wrappers  = [  ]
 
 [dependencies]
-enumflags2 = "0.7.7"
-serde = "1.0"
-static_assertions = "1.1.0"
-zbus-lockstep = { version = "0.5.0" }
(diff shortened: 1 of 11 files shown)


luukvanderduim:
@DataTriny 
Could you take a look at this, look whether the default values are okay, especially of `Value` / `OwnedValue`. I changed them to `u32` to be the same across `EventBodyOwned`, `EventBodyBorrowed`, `EventBodyQT` and `EventBodyQTBorrowed`. Size does not matter because `Value` itself is rather large. I am not particularly opinionated on the default other than we need one.

DataTriny:
I think this will perfectly fit AccessKit's use case!

The defaults look sane to me. I think C atspi puts 0i32 or "" as default value for any_data but this should not matter.

I think you should not mention the properties field in the documentation of the various clone methods, as a provider I can't put a file descriptor in there, and if a client receives one then it will simply be discarded during deserialization.

DataTriny:
Also, I'm not a native english speaker, but EventBodyBorrow sounds a bit weird to me. Shouldn't it be EventBodyBorrowed?

luukvanderduim:
> I think this will perfectly fit AccessKit's use case!
> 
> The defaults look sane to me. I think C atspi puts 0i32 or "" as default value for any_data but this should not matter.
> 
> I think you should not mention the properties field in the documentation of the various clone methods, as a provider I can't put a file descriptor in there, and if a client receives one then it will simply be discarded during deserialization.

I am curious why it is you could not put a file descriptor in a Value? I understand the protocol doesn't describe ever sending one but theoretically, why couldn't you? And other, possibly private implementations could too - or am I missing something?

luukvanderduim:
> Also, I'm not a native english speaker, but EventBodyBorrow sounds a bit weird to me. Shouldn't it be EventBodyBorrowed?

You are right. In addition to that, I don't particularly like `EventBodyQT` for the owned Qt-type either, it should be `EventBodyQtOwned` for consistency. `EventBodyOwned`, `EventBodyBorrowed`, `EventBodyQtOwned``EventBodyQtBorrowed`

luukvanderduim:
I'll go in and do those renames.
One more thing. Do we want the field "kind" renamed "type"? I see we've been a bit inconsistent in that department too but I have never known why we did that. The field is called "type" elsewhere?

DataTriny:
> I am curious why it is you could not put a file descriptor in a Value?

I never said that. I was talking about the fact that the field named `properties` of type `Properties` is mentioned in the documentation of the clone methods. Since `Properties` is an unit struct, there is no way to put a file descriptor in there and cause a panic, hence it should not be mentioned, just like it is not mentioned in the documentation for `EventBodyBorrow::to_fully_owned`.

> Do we want the field "kind" renamed "type"?

Well, `type` is a reserved keyword so we would have to use `r#type` I guess. It's probably simpler to just keep `kind`.

luukvanderduim:
> > I am curious why it is you could not put a file descriptor in a Value?
> 
> I never said that. I was talking about the fact that the field named `properties` of type `Properties` is mentioned in the documentation of the clone methods. Since `Properties` is an unit struct, there is no way to put a file descriptor in there and cause a panic, hence it should not be mentioned, just like it is not mentioned in the documentation for `EventBodyBorrow::to_fully_owned`.
> 

Aah. right! Now I understand your point. Thanks! Will adjust accordingly.

> > Do we want the field "kind" renamed "type"?
> 
> Well, `type` is a reserved keyword so we would have to use `r#type` I guess. It's probably simpler to just keep `kind`.

I understand that `type` is a keyword and hence the rename, but in one body type we do and the other we don't. I'll make it consistent, 'type' it is.

DataTriny (approved):
This can be merged if there is nothing more to add here. Thanks.

TTWNO:
@DataTriny I assume you meant to comment here when I closed the other issue.

You absolutely still can place `State` into the `Body*` types, simply use `state.into()` to convert it to a `Value/OwnedValue` respectively. We do not currently have an `Into<Value>` for these, but we could (and I have one on a test branch). Would you like me to submit a PR for that as well?

PR #224, opened 2025-02-26
`State` gains ` to_static_str()` method

luukvanderduim:
This introduces `to_static_str` for `State`.

This addresses the need for fast, non-allocating strings from a `State`,
without the risk of method implementation and enum drifting out of sync without
anyone noticing.

On efficiency of match statements on enums:  rustc is smart and (under these circumstances) compiles a jump table.

"Matches on enums with undeclared discriminants (such that the compiler can choose sequential values) are optimized down to a jump table ..."

[match statement efficiency](https://users.rust-lang.org/t/match-statement-efficiency/4488/2)

### DIFF
--- a/atspi-common/src/events/event_body.rs
+++ b/atspi-common/src/events/event_body.rs
@@ -38,7 +38,7 @@ impl Clone for EventBodyQtOwned {
 	/// This implementation of [`Clone`] *can panic!* although chances are slim.
 	///
 	/// If the following conditions are met:
-	/// 1. the `any_data` or `properties` field contain an [`std::os::fd::OwnedFd`] type, and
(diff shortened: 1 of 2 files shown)


DataTriny (approved):
Thanks Luuk, this is exactly what was missing for me.

DataTriny:
Looks like nightly clippy crashed though. Should we ignore it?

DataTriny:
I have AccessKit compiling using this branch. I still need to perform some tests but I think I have everything needed from these crates now.

TTWNO:
Already done :)

DataTriny:
Many thanks!

PR #227, opened 2025-03-10
dev(toml): Make `Cargo.toml` format pretty independent of editor.

luukvanderduim:
Make `Cargo.toml` format the same regardless of editor in use.

VSCode uses the 'Better TOML' plugin to highlight and format Cargo.toml files.

Getting the same results from eg. 'zed' requires the configuration to be independant from a local extension's settings.

This `taplo.toml` alphabetically orders entries in any 'dependencies' section, but leaves other sections 'as is'.
Also does indentation and alignment.

See [Taplo](https://taplo.tamasfe.dev/)
See [Taplo configuration](https://taplo.tamasfe.dev/configuration/file.html)

Support in editors:

zed, through 'TOML' extension (\*)
vscode: through 'Better TOML' extension (\*)
NeoVim / Vim: [taplo](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/configs/taplo.lua)

*: confirmed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,8 +107,12 @@ jobs:
         with:
           toolchain: stable
           components: rustfmt
-      - name: Run formatter
+      - name: Install Taplo
+        uses: taiki-e/install-action@taplo
+      - name: Run formatter (Rust)
         run: cargo fmt --all --check
+      - name: Run formatter (TOML)
+        run: taplo fmt --config taplo.toml --check
   rustdoc:
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


TTWNO:
Oh that's why all that funny spacing came out of nowhere, haha!

luukvanderduim:
I may do that if I can get vscode and zed behave the same with the same taplo configuration file. Because there are three parts at play, the vscode plugin, the zed plugin and the cli language server, I don't really know who is right and who is wrong.

DataTriny:
Would it make sense to set `indent_string` to a tab so that toml files are indented like the rest of the code base?

PR #230, opened 2025-03-15
zero-copy event bodies for AT-SPI2 events

luukvanderduim:
# zero-copy atspi

## zero-copy
Considering we go from `serialized body` -> `EventBodyOwned` immediately convert it to a user facing event type, 
And considering that for most events we don't even require data from the AT-SPI2 event body, we suspected that instantiating `EventBodyOwned` was a waste in most cases.

This PR eliminates the instantiation of an owned `EventBodyOwned` in favor of `EventBody<'_>` which can be either borrowed or owned:

When deserialized, it is zero-copy deserialized into its borrowed variant.
Occasionally there may be a need to have an ownedbody and for this, conversion to the other variant is offered - all
while still returning `EventBody` thus not breaking the promise made by the signature. See for exampe `MessageConversion::body`, which requires to return an owned type, but still wants to return `Self::Body`.

## passing `Header`
We also found a few other opportunities to make the path from `Message` -> `Event` more efficient as we realized that the message header was consulted multiple times in that path. Instead of creating it anew, we now pass it along from the first time it is instantiated.

compared to main these approaches yielded an appreciable performance uplift. On the AMD 5700G, 1000 messages were previously handled in ~580 microseconds where this commit reduces this to ~320 microseconds.

##  A lighter atspi with wrappers

Main had already picked up another big change, which moved all wrapping enum types in their own module and gate them behind the "wrappers" feature. Users who do not require event processing can now choose not to have it, which saves about 2.5k LoC.

## draft

Rebasing to main was not as smooth as I had wanted it to be. To be honest it was a bit messy. Therefore  I decided to make this a draft PR first so we can check if everything is in its place, kick the tires, review, test and solve problems we may find.


atspi-common/src/macros.rs
```diff
@@ -432,41 +450,49 @@ macro_rules! zbus_message_qtspi_test_case {
       #[cfg(feature = "zbus")]
      #[test]
     fn zbus_message_conversion_qtspi() {
-      // in the case that the body type is EventBodyOwned, we need to also check successful
+		use crate::events::EventTypeProperties;
+		use crate::events::MessageConversion;
+
+		// in the case that the body type is EventBodyOwned, we need to also check successful
       // conversion from a QSPI-style body.
```
TTWNO, line 457: Why were all of these missing proper formatting before? Or is the diff messed up due to a rebasing problem?

atspi-common/src/macros.rs
```diff
@@ -526,6 +552,8 @@ macro_rules! zbus_message_test_case {
 		#[cfg(feature = "zbus")]
 		#[test]
 		fn zbus_msg_conversion_validated_message_with_body() -> () {
+			use crate::events::MessageConversion;
```
TTWNO, line 555: Local, absolute imports instead of conditionally enabling imports in `event_test_cases!`. I like it!

atspi-common/src/object_match.rs
```diff
@@ -61,13 +61,16 @@ pub struct ObjectMatchRule {
 /// <https://gitlab.gnome.org/federico/at-spi2-core/-/commit/4885efedeef71e0df8210622771a0b1bb10e194d>
 impl Type for ObjectMatchRule {
 	const SIGNATURE: &'static Signature = &Signature::static_structure(&[
-		<Vec<i32>>::SIGNATURE,
```
TTWNO, line 64: Why do you prefer the explicit typing here? I specifically used the associated const `SIGNATURE` of other types because it was more readable.

atspi-common/src/object_ref.rs
```diff
@@ -21,75 +23,112 @@ pub struct ObjectRef {
...
+		let path = OwnedObjectPath::from(self.path.clone());
+		ObjectRef { name, path }
+	}
+
+	/// Create a static `ObjectRefBorrowed`, unchecked, from static string values.
+	///
+	/// # Safety
+	/// The caller must ensure that the strings are valid.
+	#[must_use]
+	pub const fn from_static_str_unchecked(sender: &'static str, path: &'static str) -> Self {
```
TTWNO, line 93: This should not be a public function.

atspi-common/src/object_ref.rs
```diff
@@ -21,75 +23,112 @@ pub struct ObjectRef {
...
-			}
-			_ => Err(zvariant::Error::IncorrectType),
-		}
+	fn try_from(value: zvariant::OwnedValue) -> Result<Self, Self::Error> {
+		let (name, path): (OwnedUniqueName, OwnedObjectPath) = value.try_into()?;
+		Ok(ObjectRef { name, path })
+	}
+}
+
+impl<'b, 'a: 'b> TryFrom<Value<'a>> for ObjectRefBorrowed<'b> {
```
TTWNO, line 127: Do we need two explicit lifetimes? Can we use a single lifetime or elide them instead?

atspi-common/src/object_ref.rs
```diff
@@ -98,3 +137,134 @@ impl From<ObjectRef> for zvariant::Structure<'_> {
...
+		use zvariant::{to_bytes, Value, LE};
+
+		let acc = ObjectRef::default();
+		let ctxt = Context::new_dbus(LE, 0);
+		let acc_value: Value<'_> = acc.into();
+		let data = to_bytes(ctxt, &acc_value).unwrap();
+		let (value, _) = data.deserialize::<Value>().unwrap();
+		let accessible: ObjectRef = value.try_into().unwrap();
+
+		assert_eq!(accessible.name.as_str(), ":0.0");
```
TTWNO, line 173: If you are going to test this, make sure the strings are `const` and used both here and in the `Default` impl. Surprise tests failing when changing a default would be irritating.

atspi-common/src/registry.rs
```diff
@@ -0,0 +1,252 @@
...
+		pub socket: ObjectRef,
+	}
+
+	impl_event_type_properties_for_event!(AvailableEvent);
+
+	event_test_cases!(AvailableEvent, Explicit);
+
+	// We cannot really register at Registry for this event, as it is emitted by the Registry itself at early startup.
+	// So, we do not have a registry event string for this event.
+	impl_member_interface_registry_string_and_match_rule_for_event!(
```
TTWNO, line 214: Good, catch. It's unconditionally emitted in `at-spi2-core/registryd/registry.c:1437`. Remove "really" in "cannot really register"; mention it's unconditional.

TTWNO (commented):
Overall, this is fantastic work! And a feat I attempted and failed multiple times—congrats! Looking forward to further performance tuning this with you.

atspi-common/src/events/document.rs
```diff
@@ -59,86 +58,108 @@ pub struct PageChangedEvent {
...
 impl_from_object_ref!(LoadCompleteEvent);
 
+impl_event_type_properties_for_event!(ReloadEvent);
 event_test_cases!(ReloadEvent);
 impl_to_dbus_message!(ReloadEvent);
 impl_from_dbus_message!(ReloadEvent);
 impl_event_properties!(ReloadEvent);
 impl_from_object_ref!(ReloadEvent);
 
+impl_event_type_properties_for_event!(LoadStoppedEvent);
```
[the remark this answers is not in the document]
luukvanderduim: The benefits are threefold 
- avoid repeating ourselves
- avoid the complexity of blanket impls.
- improving the ability for tooling to see what concrete type is dealt with, instead of 'some T'

atspi-common/src/events/mouse.rs
```diff
@@ -36,110 +38,145 @@ pub struct ButtonEvent {
...
-	const MATCH_RULE_STRING: &'static str =
-		"type='signal',interface='org.a11y.atspi.Event.Mouse',member='Rel'";
-	const REGISTRY_EVENT_STRING: &'static str = "Mouse:";
-}
-
 #[cfg(feature = "zbus")]
-impl MessageConversion for RelEvent {
-	type Body = EventBodyOwned;
+impl MessageConversion<'_> for RelEvent {
+	type Body<'a> = EventBody<'a>;
```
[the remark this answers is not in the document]
luukvanderduim: 'progress toward' as in 'preparatory work, but not having reached yet'?
This does deserialize to it's borrowed variant and also, being an enum of both borrowed and owned, this type satisfies the requirement in `MessageConversion` to both deserialize to `Self::Body` return `Self::Body` without bounding it to some lifetime.

atspi-common/src/object_ref.rs
luukvanderduim: You are right. We can do with one.
I tried hard to find an example where the supposed flexibility of two lifetimes was beneficial but could not find any and I currently suspect this is because we are not tied to `Value<'a>` but tied to whatever it was pointing to. Lifetimes are hard, and I am easily misguided still.

atspi-common/src/events/mouse.rs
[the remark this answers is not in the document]
TTWNO: > 'progress toward' as in 'preparatory work, but not having reached yet'?

There are a couple body types which are still unborrowed; I'm thinking `CacheIem`, `LegacyCacheItem`, `EventListeners` and `ObjectRef`. Everything that uses `EventBody<'_>` is indeed now borrowed.

luukvanderduim (commented):
~I _think_ I addressed all questions you had in this round.~

atspi-common/src/events/traits.rs
```diff
@@ -43,89 +44,69 @@ pub trait EventProperties {
...
+	/// Will fail if no matching member or body signature is found.
+	fn try_from_message_interface_checked(
+		msg: &zbus::Message,
+		hdr: &Header,
+	) -> Result<Self, AtspiError>
+	where
+		Self: Sized;
+}
+
+// TODO: Document why this can't be `TryFrom<&zbus::Message>`.
```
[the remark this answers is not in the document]
luukvanderduim: Removing it will be a simple exercise. I already have the `TryFrom`s implemented in `event_wrappers`, only their implementation needs to be changed:

```Rust 
#[cfg(feature = "zbus")]
impl TryFrom<&zbus::Message> for WindowEvents {
	type Error = AtspiError;
	fn try_from(msg: &zbus::Message) -> Result<Self, Self::Error> {
		Self::try_from_message(msg)
	}
}
```

atspi-common/src/events/traits.rs
```diff
@@ -43,89 +44,69 @@ pub trait EventProperties {
...
-/// - `DBus` interface name
-/// - `DBus` match string: used to tell `DBus` you are interested in a particular signal
-/// - accessibility registry event string: used to tell the accessibility registry that you are interested in a particular event
-///
-/// This trait *is not* object-safe.
-/// For a similar, but object-safe trait, see [`EventProperties`].
-pub trait BusProperties {
-	/// The `DBus` member for the event.
-	/// For example, for an [`crate::events::object::TextChangedEvent`] this should be `"TextChanged"`
+assert_obj_safe!(EventTypeProperties);
```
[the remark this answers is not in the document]
luukvanderduim: I don't know.
Importing macros has changed a bit over the years. But I hadn't heard about the 'no need to even name' feature. 

But seriously, this is weird but it is in scope, rust-analyzer sees it and rustc does not complain.
Importing the macro 2018 style (single macro per use statement) gives an unused import warning.

atspi-common/src/events/traits.rs
```diff
@@ -140,17 +121,18 @@ where
...
+
 	/// Validate the interface string via [`zbus::message::Header::interface`] against `Self`'s assignment of [`BusProperties::DBUS_INTERFACE`]
 	///
 	/// # Errors
 	///
 	/// - [`type@AtspiError::MissingInterface`] if there is no interface
 	/// - [`type@AtspiError::InterfaceMatch`] if the interfaces do not match
-	fn validate_interface(msg: &zbus::Message) -> Result<(), AtspiError> {
-		let header = msg.header();
+	#[inline]
```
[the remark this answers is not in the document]
luukvanderduim: Great catch, Thank you!

Silly experiment from a week ago that I forgot to remove.

luukvanderduim:
> Overall, this is fantastic work! And a feat I attempted and failed multiple times—congrats! Looking forward to further performance tuning this with you.

Thank you Tait.

atspi-common/src/events/object.rs
```diff
@@ -895,6 +1116,46 @@ impl_event_properties!(TextSelectionChangedEvent);
 impl_from_object_ref!(TextSelectionChangedEvent);
 
 event_test_cases!(TextChangedEvent);
+
```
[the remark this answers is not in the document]
TTWNO: But why is there a separate inner module for `TextChangedEvent`?

atspi-common/src/macros.rs
luukvanderduim: Macros are always messy. For some reason unknown to me, cargo-fmt does not touch macros.
I just either clean them up manually a bit - or when my fmt-OCD really kicks in - I go somewhere the macro is used, expand it there, have it formatted, copy it back and put the `$` variables back in place.

TTWNO:
If possible, move what needs to be from `src/registry.rs` to `src/events/registry.rs`

TTWNO:
I see that `events/mod.rs` is very minimal now. This is fantastic since that file was getting to be quite the monstor :D

luukvanderduim:
> If possible, move what needs to be from `src/registry.rs` to `src/events/registry.rs`

Done!

luukvanderduim:
I noticed I have not gated the `event_wrappers` module in `events/mod.rs`.

fix incoming

luukvanderduim:
@TTWNO  Do we build examples in CI?

TTWNO:
No I don't think so.

TTWNO (approved):
LGTM! I've filed the only issue I want us to fix later and call it here.

Let's.... do our best to make smaller PRs in the future; to be fair this is not your fault as this basically re-writes a lot of our library.

TTWNO:
I would like you to squash and merge, detail the summary of changes without overwhelming future log readers of all the finicky details of what you did here.

Awesome work! Let's bring 'er home and then we can focus on other things!

luukvanderduim:
> Let's.... do our best to make smaller PRs in the future; to be fair this is not your fault as this basically re-writes a lot of our library.

Wholeheartedly agree. There was avoidable pain here. However, some of our new ideas and insights about  how we could improve performance by deferring deserialization and doing that lazily for all events for instance, has impact almost all over. So big PRs will remain, I'm afraid.

Here are some observations I've made along the way.

- It felt really good doing 'unbounded' work in a branch. Why is that? 
 I think it is because of a short effort-reward feedback loop.
Also, this PR was all about making things go 'blazingly brrr', which we just love. 

- I would like us to maximize joy and avoid the problems I created along the way as much we can.

- I should not let branches diverge again - no one should of course, but this one was on me. 
It's one thing having a big PR, requiring conflict resolving from changes coming from the feature branch, but when removals and additions keep coming from either direction, things get hairy. 
Best is to rebase every time `main` gets a change and not kick that can down the road.

- Secondly I would like us to (try to) keep changes confined to single purpose aka atomic commits. 
Such that in the context of a big PR, we at least get to reorder commits in interactive rebase to make a rebase easier.
So best practice is to - even in a big PR - let a commit do one thing only. A commit should do only what the description says it does.

PR #234, opened 2025-03-22
doc: Use '\' instead of double spaces for hard line breaks in docs.

luukvanderduim:
In docs we have been using double spaces to indicate a hard line break. Clippy does not approve of this practice.
Clippy suggests using '\' instead.


TTWNO:
Since this is now a nightly lint, it was caught and fixed as part of #230

PR #236, opened 2025-03-31
build(examples): Fixups for example builds

luukvanderduim:
Fixed minor import path changed.

Disabled `required-features = ["feat", "feat"]` on all three examples because it does not seem to matter - which is probalby indicative of failing gates. Needs attention.


TTWNO:
You are right. The examples import `atspi`, not any sub-libraries. Good catch!

PR #241, opened 2025-04-28
fix(proxies): disable property caching by default

luukvanderduim:
By annotating property methods in proxy 'trait's with:

```rust
#[zbus(property(emits_changed_signal = "false"))]
fn property_getter()
```

disables the property cache because zbus assumes caching properties is only safe if it can listen for updates on properties. If property changes are not  (guaranteed) to be transmitted, zbus cannot know whether the data it has in cache has gone stale.

[zbus v5.5.0 docs on `emits_changed_signal` arguments:](https://docs.rs/zbus/latest/zbus/attr.proxy.html)
```text
"false" - change signal is not (guaranteed to be) emitted if the property changes. 
This disables property value caching, and does not generate a listener method for the change signal.
```

### DIFF
--- a/atspi-proxies/src/accessible.rs
+++ b/atspi-proxies/src/accessible.rs
@@ -162,11 +162,11 @@ pub trait Accessible {
 	/// a test program may have to recursively get the children to find a specific id.
 	/// This is because accessible objects can be created dynamically, and they do not always
 	/// correspond to a static view of an application's data.
-	#[zbus(property)]
+	#[zbus(property(emits_changed_signal = "false"))]
 	fn accessible_id(&self) -> zbus::Result<String>;
 
 	/// Number of accessible children for the current object.
-	#[zbus(property)]
+	#[zbus(property(emits_changed_signal = "false"))]
 	fn child_count(&self) -> zbus::Result<i32>;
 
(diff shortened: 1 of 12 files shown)


TTWNO:
Why do we want this exactly?

luukvanderduim:
> Why do we want this exactly?

To remove a foot gun.

Without it, if you create a proxy but do not opt-out of property caching, obtaining properties will (likely) error.
With this, property caching will be disabled by default so building proxies will be more robust.

TTWNO:
@luukvanderduim AccessKit uses the functionality of the change emitted signals. This will have to be reverted.

Is there any other way we can still allow for the changed signals bus disable caching?

luukvanderduim:
Property signals are sent regardless of us wanting cached properties. 
My educated guess is AccessKit was using the generated stream? Is there an issue associated with this problem?
One could try and register for `PropertyChanged` signals and handle them as you do..

TTWNO:
Yes, they depend on the generatwd stream. It would be nice of the stream could be generated but also force non-caching on the zbus side

luukvanderduim:
I found no issue associated with this problem, how did you learn about it?

PR #246, opened 2025-05-06
Document all Object events

TTWNO:


TTWNO:
cc @luukvanderduim

luukvanderduim (changes requested):
I like the references where they apply. 
I like the helpful texts, like the ones describing how to get the actual attributes etc.
The self-referencing docs, where the identifier is copied to the docs, are not as helpful.
I understand that writing good documentation is a lot of tedium and time consuming.
But when you enable the lint on docs and then we quickly fill in the docs with quick-n-dirty placeholders we are essentially worse off. A Clippy run won't tell us where it hurts. So in that respect Maybe no documentation is preferable over poor placeholders?

PR #248, opened 2025-05-19
proxies: fix metadata to match other crates (include license files)

decathorpe:
I am working on packaging the atspi* crates for Fedora Linux as a dependency of `elevado` (and later, potentially odilia). I noticed that while the `atspi-proxies` directory in the git repo contains the license texts, they are not included in published crates because the `package.include` setting in the `atspi-proxies` workspace member doesn't match the same setting in other workspace members and doesn't list `LICENSE-*` files as to be included. This seems to be a simple oversight, which this PR fixes. It would be great if you could publish a new version with these changes (0.10.2?) soon-ish. :)


TTWNO:
Good find. Thank you; we appreciate it :)

Release will be made in the next 10 minutes.

decathorpe:
Awesome, thank you! 🚀

TTWNO:
@decathorpe done!

PR #253, opened 2025-05-27
async io clarification

TTWNO:
- **Clarify async-io/tokio features**
- **Add cargo-hack to CI for feature unification checks**

PR #254, opened 2025-05-29
Fix Property Names

TTWNO:
- **Use property names from at-spi2-core; add tests for coverage**

### DIFF
--- a/atspi-common/src/events/object.rs
+++ b/atspi-common/src/events/object.rs
@@ -124,68 +124,64 @@ impl TryFrom<EventBody<'_>> for Property {
 		let property = body.kind();
 
 		match property {
-			a if a == ACCESSIBLE_NAME_PROPERTY_NAME => Ok(Self::Name(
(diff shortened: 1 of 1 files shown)


TTWNO:
Depends on #253

atspi-common/src/events/object.rs
```diff
@@ -109,17 +124,17 @@ impl TryFrom<EventBody<'_>> for Property {
 		let property = body.kind();
 
 		match property {
-			"accessible-name" => Ok(Self::Name(
+			a if a == ACCESSIBLE_NAME_PROPERTY_NAME => Ok(Self::Name(
```
DataTriny, line 127:
```suggestion
			ACCESSIBLE_NAME_PROPERTY_NAME => Ok(Self::Name(
```

Please also update all match arms below.

DataTriny (changes requested):

DataTriny (approved):

PR #262, opened 2025-06-07
Add example to watch and print selected text

C-Loftus:
I think it would be useful to have an example to show how to use the text proxy and how to monitor the currently selected text.

### Nautilus path issue

I did run into one issue with the example, namely in Nautilus if I try to select the text for a file path, it sometimes gives this error. 

```
Error: MethodError(OwnedErrorName("org.freedesktop.DBus.Error.InvalidArgs"), Some("Not a valid selection: 0"), Msg { type: Error, serial: 116, sender: UniqueName(":1.87"), reply-serial: 304, body: Str, fds: [] })
```

Otherwise I have tested this and it works well in Firefox.

### More than one selection support

If I try to support more than 1 range it seems like there is a dbus issue
```rs
                    let ranges = text_proxy.get_nselections().await?;
                    for i in 0..ranges {
                        let (start, end) = text_proxy.get_selection(i).await?;
                        println!("{}", text_proxy.get_text(start, end).await?);
                    }
```

```
host@computer ~/g/cAtspi (watchTextSelectionChanges) [1]> cargo run --example selected-text
   Compiling atspi v0.27.0 (/home/host/github/cAtspi/atspi)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.90s
     Running `target/debug/examples/selected-text`
Error: MethodError(OwnedErrorName("org.freedesktop.DBus.Error.UnknownMethod"), Some("Method \"GetNselections\" with signature \"\" on interface \"org.a11y.atspi.Text\" doesn't exist\n"), Msg { type: Error, serial: 229106, sender: UniqueName(":1.14"), reply-serial: 7, body: Str, fds: [] })
```

### DIFF
--- a/atspi/examples/selected-text.rs
+++ b/atspi/examples/selected-text.rs
@@ -1,0 +1,36 @@
+//! This example demonstrates how to get the currently selected text.
+//!
+//! ```sh
+//! cargo run --example selected-text
+//! ```
+//! Authors:
+//!    Colton Loftus
+
+use atspi::{events::object::{TextSelectionChangedEvent}, ObjectEvents};
+use atspi_proxies::{accessible::ObjectRefExt, proxy_ext::ProxyExt};
+use futures_lite::stream::StreamExt;
+use std::{error::Error};
+
+#[async_std::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+	let atspi = atspi::AccessibilityConnection::new().await?;
+    let conn = atspi.connection();
+	atspi.register_event::<ObjectEvents>().await?;
+
+	let mut events = atspi.event_stream();
+
+	while let Some(ev) = events.next().await {
+        match ev {
+            Ok(ev) => {
+                if let Ok(ev) = <TextSelectionChangedEvent>::try_from(ev) {
+                    let text_proxy = ev.item.into_accessible_proxy(conn).await?.proxies().await?.text().await?;
+                    const ASSUME_ONLY_ONE_SELECTED_RANGE: i32 = 0;
+                    let (start, end) = text_proxy.get_selection(ASSUME_ONLY_ONE_SELECTED_RANGE).await?;
+                    println!("{}", text_proxy.get_text(start, end).await?);
+                }
+            }
+            Err(err) => eprintln!("Error: {err}"),
+        }
+	}
+	Ok(())
+}
--- a/atspi/examples/selected-text.rs
+++ b/atspi/examples/selected-text.rs
@@ -6,31 +6,44 @@
 //! Authors:
 //!    Colton Loftus
 
-use atspi::{events::object::{TextSelectionChangedEvent}, ObjectEvents};
+use atspi::{events::object::TextSelectionChangedEvent, ObjectEvents};
 use atspi_proxies::{accessible::ObjectRefExt, proxy_ext::ProxyExt};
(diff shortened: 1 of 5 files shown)


atspi/examples/selected-text.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use atspi::{events::object::{TextSelectionChangedEvent}, ObjectEvents};
+use atspi_proxies::{accessible::ObjectRefExt, proxy_ext::ProxyExt};
+use futures_lite::stream::StreamExt;
+use std::{error::Error};
+
+#[async_std::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+	let atspi = atspi::AccessibilityConnection::new().await?;
+    let conn = atspi.connection();
```
luukvanderduim, line 17: Just run cargo fmt to get these alignments right.

atspi/examples/selected-text.rs
```diff
@@ -0,0 +1,36 @@
...
+	atspi.register_event::<ObjectEvents>().await?;
+
+	let mut events = atspi.event_stream();
+
+	while let Some(ev) = events.next().await {
+        match ev {
+            Ok(ev) => {
+                if let Ok(ev) = <TextSelectionChangedEvent>::try_from(ev) {
+                    let text_proxy = ev.item.into_accessible_proxy(conn).await?.proxies().await?.text().await?;
+                    const ASSUME_ONLY_ONE_SELECTED_RANGE: i32 = 0;
```
luukvanderduim, line 27: Please move the `const` up, just below the `use` statements.

atspi/examples/selected-text.rs
```diff
@@ -0,0 +1,36 @@
...
+                    let (start, end) = text_proxy.get_selection(ASSUME_ONLY_ONE_SELECTED_RANGE).await?;
```
luukvanderduim, line 28: Where did you find the meaning of the argument? Just curious. If it is a credible source, let's document it on our method.

luukvanderduim (changes requested):
I ran it and found it is a delightful little example!

I can somewhat infer from the `const`'s name what the definition of the argument should be "If the argument is zero we expect only one selection"? 

If you move the declaration of the const up, we're good to go as far as I am concerned.

Thanks a lot!

C-Loftus:
Believe I have tidied that up?

```
host@computer ~/g/cAtspi (watchTextSelectionChanges)> cargo fmt
host@computer ~/g/cAtspi (watchTextSelectionChanges)> cargo clippy --examples --tests --no-deps --all-features -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
```

atspi/examples/selected-text.rs
luukvanderduim: Where does it say that 0 indicates 'only one selection'?

luukvanderduim: So it is an index?

TTWNO:
@C-Loftus Can you please rebasr this [with signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

C-Loftus:
> @C-Loftus Can you please rebasr this [with signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

I believe I have this signed and verified now.

PR #265, opened 2025-06-14
Add a registry root method

C-Loftus:
- adds a `get_registry_root` method to make tree traversals easier to do for those who are less familiar with dbus. 
    - ideally i think it would be helpful for specific dbus destinations / interfaces to be abstracted away
- changed luuk's tree traversal example to use this
- feel free to deny if not relevant

### DIFF
--- a/atspi-connection/src/lib.rs
+++ b/atspi-connection/src/lib.rs
@@ -333,6 +333,8 @@ impl AccessibilityConnection {
 	/// # tokio_test::block_on(async {
 	/// let connection = atspi_connection::AccessibilityConnection::new().await.unwrap();
 	/// let root = connection.root_accessible_on_registry().await.unwrap();
+	/// let children = root.get_children().await;
+	/// assert!(children.is_ok());
 	/// # });
 	/// ```
 	/// # Errors


C-Loftus:
> Lets fix up the loose ends and get this in!

Will do!

Could you let me know where, if anywhere, I should put tests for this? I didn't see any tests in the same file.

> How about a method on `ObjectRef` that can returns an `ObjectRef` for the root object given something that can become a `BusName`?

That seems reasonable; just so I am clear that could be another follow up PR right? Not necessarily mutually exclusive?

My thought was that a tree traversal from the root of the registry is a very common operation for some assistive technology workflows. A lot of folks don't have any experience with dbus and it seems like having a direct helper for this (or any other very common dbus query) makes it a bit easier to get started.  I think many users may be unaware of what bus name to use in many situations

luukvanderduim:
> > Lets fix up the loose ends and get this in!
> 
> Will do!
> 
> Could you let me know where, if anywhere, I should put tests for this? I didn't see any tests in the same file.
> 

In this particular case you can test in the docs by writing an example wherein you `assert!` that indeed the call to `get_children` returns `Ok(_)`.
That being said, it is difficult to test a proxy or a stream because you need an implementer of that interface on the bus or a program that emits the stream in order to receive it.
We could put in the effort to mock a stream or a program that exposes an interface but that would be just one implementation. What about Qt, Gtk3, Gtk4, AccessKit, Gecko, Chromium?
We have the XML protocol descriptions and we consider it our point of reference. We try to follow it and assume the others do too. 
Ideally we have [Rust integration tests](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) where we fire up in instance of each provider and elicit the same event from each and see whether they do what we (and the specs) expect. But aside from a lot of work, I would not know how in many cases either. 
Note that doctests are integration tests too  as they cannot touch private items and are compiled as their own crate. 

> > How about a method on `ObjectRef` that can returns an `ObjectRef` for the root object given something that can become a `BusName`?
> 
> That seems reasonable; just so I am clear that could be another follow up PR right? Not necessarily mutually exclusive?

Yes.

> 
> My thought was that a tree traversal from the root of the registry is a very common operation for some assistive technology workflows. A lot of folks don't have any experience with dbus and it seems like having a direct helper for this (or any other very common dbus query) makes it a bit easier to get started. I think many users may be unaware of what bus name to use in many situations

I'll argue listening is perhaps more common than tree traversing. Keep in mind that traversing trees is relatively expensive because you do a lot of bus communications and traversing can only tell you about the state of the UI, user- or machine interaction (change of state) comes as events.
If you are writing some inspection tool you cannot do without traversing trees but if you write a screen reader your main objective is to listen. Maybe you build a tree once (cache) update it accordingly as relevant event happen but otherwise you listen for events and generate speech.

luukvanderduim (approved):

PR #267, opened 2025-06-17
Feature: Peer-to-Peer for atspi.

luukvanderduim:
# P2P

The `Application` interface offers `GetApplciationBusAddress` which allows ATs to obtain a bus address for direct communication with many applications but with lower latency.

This PR proposes to extend `AccessibilityConnection` with P2P capabilities.

If the feature "p2p" is enabled, the `AccessibilityConnection` will be assembled with a list of current applications that report to support P2P communication.
A task will also be spawned to continuously listen for new applications entering the bus or applications leaving the bus by
listening for `NameAcquired` or `NameLost` signals emitted by the `DBus` daemon. 

## trait `P2P` and `Peer`

A list of `Peer`s is kept by the `AccessibilityConnection`. A `Peer` contains the application's `BusName` (known to always be unique), socket address and the zbus::Connection to the application.

The `P2P` trait offers various ways to obtain a `Peer`, get `AccessibleProxy` for a given `ObjectRef` or a root `AccessibilityProxy` for a given path. Users can also obtain `Proxies` objects for a given `ObjectRef` to quickly access any interface an object says to implement.
 
## Example `p2p_tree`

`p2p_tree` compares common bus method calls with two ways of performing peer-to-peer communication when building a tree of all accessible applications combined.

```shell
     Running `target/release/examples/p2p_tree`
Number of accessible applications on the a11y-bus: 22

D-Bus operation                Node count Time (ms)
----------------------------------------------------
Building tree (bus)            17138      3673.46
Building tree (P2P)            17138      1862.57
Building tree (P2P parallel)   17138      1776.78
----------------------------------------------------
Speedup (p2p vs bus)           1.97x
Speedup (p2p-par vs bus)       2.07x
```


atspi-connection/p2p.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+The example walks all nodes in the accessibility tree and queries each node for its role. The example registers the amount of time the operation took.
+
+It does so in three ways:
+The `p2p_tree` example demonstrates three ways to traverse the accessibility tree and measure performance:
+
+1. **Bus:** Queries all nodes of all applications over the accessibility bus.
+2. **P2P Sequential:** Queries each application's nodes over a P2P connection (if available), one after another.
+3. **P2P Parallel:** Queries each application's nodes over P2P connections (if available) concurrently, leveraging parallelism.
```
TTWNO, line 18: Parallelism and concurrency are different. I'd recommend not interweaving the two terms casually.

Parallelism refers to doing things _at the same time_ (i.e., using multi-processing), concurrency is about doing things _while waiting for other things to finish_ (i.e., check user interactions while waiting for an I/O stream to finish reading).

Even if your example used parallel processing (it doesn't), the application on the other side likely does not. The suffix `_concurrent` or similar fits better than a Rayon-like `_par`.

atspi-connection/p2p.md
```diff
@@ -0,0 +1,43 @@
...
+| Building tree (bus)          | 6386       | 1291.77   |
+| Building tree (P2P)          | 6386       | 688.18    |
+| Building tree (P2P parallel) | 6386       | 554.23    |
+|------------------------------|------------|-----------|
+| Speedup (p2p vs bus)         |            | 1.88x     |
+| Speedup (p2p-par vs bus)     |            | 2.33x     |
+
+* Tested on Intel Core Ultra 7 155H*
+
+The actual latency improvement depends on your hardware. Faster, more IO-bound systems tend to benefit more from P2P, as they spend less time waiting for the bus and more time processing requests in parallel.
```
TTWNO, line 33: Again, please remove the parallel language. `atspi` does not offer _any_ parallelism.

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,746 @@
...
+
+use crate::AtspiResult;
+
+const ACCESSIBLE_ROOT_OBJECT_PATH: &str = "/org/a11y/atspi/accessible/root";
+const REGISTRY_WELL_KNOWN_NAME: &str = "org.a11y.atspi.Registry";
+
+/// Represents a peer with the name, path and connection for the P2P peer.
+#[derive(Clone, Debug)]
+pub struct Peer {
+	unique_name: OwnedUniqueName,
```
TTWNO, line 33: You can use `zbus::BusName` here. No need for the two fields.

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,746 @@
+//! Extends the `AccessibilityConnection` with P2P capabilities.
```
TTWNO, line 1: You explanations and commentary throughout the file are a pleasure to read! I wish all code was documented this well—including my own.

Thank you for the nice read this afternoon!

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,746 @@
...
+	Address,
+};
+
+#[cfg(feature = "tracing")]
+use tracing::{debug, info, warn};
+
+use crate::AtspiResult;
+
+const ACCESSIBLE_ROOT_OBJECT_PATH: &str = "/org/a11y/atspi/accessible/root";
+const REGISTRY_WELL_KNOWN_NAME: &str = "org.a11y.atspi.Registry";
```
TTWNO, line 28: See #270

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,746 @@
...
+	async fn remove_unique(&self, unique_name: &OwnedUniqueName) -> AtspiResult<()> {
+		let mut peers = self.peers.lock().await;
+		peers.retain(|peer| peer.unique_name() != unique_name);
+		Ok(())
+	}
+
+	/// Inserts a new `Peer` with a well-known name into the list of peers.
+	async fn insert_well_known(
+		&self,
+		well_known_name: OwnedWellKnownName,
```
TTWNO, line 184: Is it true that there can be many names for one application's accessibility interface?

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,746 @@
...
+				let mut name_owner_changed_stream = match dbus_proxy.receive_name_owner_changed().await {
+					Ok(stream) => stream,
+					Err(_err) => {
+						#[cfg(feature = "tracing")]
+						warn!("Failed to get DBusProxy `NameOwnerChanged` stream: {}", _err);
+						return;
+					}
+				};
+
+				while let Some(name_owner_event) = name_owner_changed_stream.next().await {
```
TTWNO, line 272: Great documentation here on how this all works. It's just a lot to read in one go.

It is _very_ deeply nested. It would be better to separate:

1. The spawned async task
2. The very long, nested match arms.

into their own functions.

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,746 @@
...
+									let _result = self.remove_unique(&OwnedUniqueName::from(unique_name)).await;
+
+									#[cfg(feature = "tracing")]
+									_result.unwrap_or_else(|err| {
+										warn!("Failed to remove unique name: {unique_name}: {err}");
+									});
+								}
+
+								// Unknown combination.
+								(_, _) => {
```
TTWNO, line 318: Why is this an unknown combination, but all cases are covered for the `WellKnwon` name?

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,746 @@
...
+	/// - The socket address cannot be parsed.
+	///
+	pub(crate) async fn try_new<B, S>(
+		bus_name: B,
+		socket: S,
+		conn: &zbus::Connection,
+	) -> Result<Self, AtspiError>
+	where
+		B: Into<OwnedBusName>,
+		S: TryInto<Address>,
```
TTWNO, line 419: Solid trait bounds here! This is much more useful than taking direct parameters.

atspi/examples/p2p-tree.rs
```diff
@@ -0,0 +1,344 @@
...
+use futures::{
+	future::{join_all, try_join_all},
+	stream::FuturesUnordered,
+};
+use zbus::{proxy::CacheProperties, Connection};
+
+type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
+
+const REGISTRY_DEST: &str = "org.a11y.atspi.Registry";
+const ROOT_ACCESSIBLE_PATH: &str = "/org/a11y/atspi/accessible/root";
```
TTWNO, line 25: See #270

atspi/examples/p2p-tree.rs
```diff
@@ -0,0 +1,344 @@
...
+		"D-Bus operation", "Node count", "Time (ms)"
+	);
+	let table_divider = "-".repeat(DESC_WIDTH + NODE_COUNT_WIDTH + TIME_WIDTH + 2).to_string();
+
+	// Building tree (bus)
+	let now = std::time::Instant::now();
+	let _tree_bus = A11yNode::from_accessible_proxy_bus(registry.clone()).await?;
+	let bus_elapsed = now.elapsed();
+	let bus_tree_line = format!(
+		"{:<DESC_WIDTH$} {:<NODE_COUNT_WIDTH$} {:<TIME_WIDTH$.2?}",
```
TTWNO, line 277: Did not know about these formatting options!

luukvanderduim: Me neither, LLM suggestion. They do formatting rather well but don't tell anyone the formatting was generated. ;)

atspi-connection/src/p2p.rs
luukvanderduim: It is a bit dbus quirkyness due to expressing all permutations with the one body type. When receiving `NameOwnerChanged`, you receive a 'name', and an 'old' and 'new' owner. When name is a `UniqueName` it only makes sense to have either a new "owner" to indicate the name appeared on the bus or an old "owner" to indicate a name left the bus.

I made a table. You can find it in the updater task in the sources too.

```markdown
	 `NameOwnerChanged` table (U = Unique, W = Well-Known):
	 | Name | Old Owner | New Owner | Operation |
	 |------|-----------|-----------|----------|
	 |   U  |   None    |  Some(U)  |  Add     |
	 |   U  |   Some(U) |    None   |  Remove  |
	 |   W  |   None    |  Some(U)  |  Add     |
	 |   W  |   Some(U) |    None   |  Remove  |
	 |   W  |   Some(U) |  Some(U)  |  Replace |

```

atspi-connection/src/p2p.rs
luukvanderduim: I've "separate concerns" here which mitigates complexity in the long match (to some degree). You may want to revisit since the last time.

TTWNO:
> I would prefer to add testing to CI using (something akin to) [test-atspi-p2p](https://github.com/luukvanderduim/test-atspi-p2p) and firing up that application, run it to completion and assert a thing or two.

DBus works fine in CI—including using `cross`, P2P, etc., so that would work.

atspi-connection/src/p2p.rs
luukvanderduim: No. This is not an 'or'-situation. The `Peer` will always have a `UniqueName` which may (or may not) own a `WellKnownName`. For a `WellKnownName' to make any snse it must have an owner - a `UniqueName`. 
In fact a `UniqueName` (a connection) can own as many `WellKnownName`'s as it wants. (We would represent that as having multiple `Peer`s). `WellKnownName`'s however cannot be on the bus on their own, they always need an owner - a process with a connection to the bus to communicate with.

TTWNO:
If we do the p2p integration test in sequence with the coverage job, I wonder if we can combine results? No worries, don't fret over it now, let's review again and get this merged first.

luukvanderduim:
> If we do the p2p integration test in sequence with the coverage job, I wonder if we can combine results? No worries, don't fret over it now, let's review again and get this merged first.

Not all tests are made equal. You can have great coverage with poor tests and vice versa. In itself coverage is not the best metric. In P2P we have a handful of doc-tests but I am not sure coverage counts those - and these integration tests which tests arguably the most important bit of the listener. That it is there, that it catches the signal, that it adds and remove supported apps.

.github/workflows/ci.yml
```diff
@@ -289,4 +289,89 @@ jobs:
       - name: Generate lockfile with minimal direct versions
         # Create a lockfile with minimal direct versions for reproducible builds.
         # See advisory: https://blog.rust-lang.org/2023/08/29/committing-lockfiles/
-        run: cargo +nightly generate-lockfile -Z direct-minimal-versions
\ No newline at end of file
+        run: cargo +nightly generate-lockfile -Z direct-minimal-versions
```
TTWNO, line 292: Fantastic work getting this functioning. Really glad you stuck through and made it happen!

.vscode/settings.json
TTWNO, line 1: Do not commit editor settings.

atspi-common/tests/tests.rs
```diff
@@ -47,10 +47,12 @@ async fn test_recv_remove_accessible() {
 
 	loop {
 		let to = events.try_next().await;
-		let event = to
-			.expect("stream timed out")
-			.expect("stream closed")
-			.expect("conversion to `Event` failed");
+		let bus_event = to.expect("stream timed out").expect("stream closed");
```
TTWNO, line 50: Good catch!

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,768 @@
...
+	/// This function is intended for use in building the initial list of peers.
+	///
+	/// If given a `UniqueName`, it will check if the peer also owns a well-known name.
+	/// If given a `WellKnownName`, it will query the D-Bus for the unique name of the peer.
+	///
+	/// # Errors
+	/// - `DBusProxy` cannot be created.
+	/// - The socket address cannot be parsed.
+	///
+	pub(crate) async fn try_new<B, S>(
```
TTWNO, line 64: Great use of generics here; learning from Zeeshan are we?

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,768 @@
...
+			}
+			BusName::WellKnown(well_known_name) => {
+				let bus_name = BusName::from(well_known_name.clone());
+				let owned_unique_name = dbus_proxy.get_name_owner(bus_name).await?;
+				let owned_well_known_name = OwnedWellKnownName::from(well_known_name.clone());
+				(owned_unique_name, Some(owned_well_known_name))
+			}
+		};
+
+		let socket_address = socket
```
TTWNO, line 133: This should be done higher up; since it is not dependent on the other portions of the function finishing correctly.

If possible, this error should be able to bubble up before any bus activity. It makes errors potentially more consistent.

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,768 @@
...
+									peers.remove_unique(&unique_name).await;
+
+									#[cfg(feature = "tracing")]
+									info!("Peer with unique name: {unique_name} left the bus - removed from peer list.");
+								}
+
+								// Unknown combination.
+								(_, _) => {
+									#[cfg(feature = "tracing")]
+									debug!("NameOwnerChanged` with unique name: {unique_name} has unknown argument combination ({old:?}, {new:?}).");
```
TTWNO, line 508: Some of these are `debug!` and some are `debug_assert!`; how do you differentiate the two cases?

rust-analyzer.json
TTWNO, line 1: Remove this file.

atspi-connection/src/p2p.rs
```diff
@@ -0,0 +1,768 @@
...
+//!
+//! When zbus users use `tokio` (zbus feature "tokio" set), zbus will latch onto the tokio runtime.
+//! The `Executor` instance will be empty and all zbus tasks are run on the user's tokio runtime.
+//! However, when using any other executor (smol, glommio, async-std, etc.), each `Connection` will spin up a thread with an `async_executor::Executor`.
+//!
+//! Typically an application will have a single connection, but with P2P, your application will have a connection with each application that supports it.
+//! Consequently, on anything but tokio, applications will get an extra thread with an `async_executor` for each connection!
+//! (So picking smol won't necessarily make your application small in the context of P2P.)
+
+use async_lock::Mutex;
```
TTWNO, line 16: You do not need [and are encouraged not to use (from tokio)](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use) async locks, where possible. There are also some functions like `remove_unique`, `peers` and a few others that do not need to me async (or are only async due to the async Mutex). Only use async when it is a hard requirement.

Anyways, you don't even hold a value across an await paint so you should be fine :)

TTWNO (changes requested):
Almost there! Thank you again for your hard work!

.vscode/settings.json
luukvanderduim: Sorry, not my intention. Good catch!
I'll remove add it to .gitignore even!

.github/workflows/ci.yml
TTWNO: Oops I mean the second section that does the tests :)

rust-analyzer.json
luukvanderduim: This one escaped, well caught!

.github/workflows/ci.yml
luukvanderduim: Ah _that_ one.. The fruits of sheer stubbornness.

atspi-connection/src/p2p.rs
luukvanderduim: Generally I am against doing work before earlier bail-outs could happen, but here I agree. Moved up.

atspi-connection/src/p2p.rs
luukvanderduim: You are right, they are not needed.
I am sure I did not put them there out of precaution. I think I refactored away it's necessity.
That's great because it simplifies code, removes async requirements on a handful of methods and drops the dependency on async-lock.

atspi-connection/README.md
```diff
@@ -10,7 +10,14 @@ See the examples folder and documentation on how to use this library.
 
 ## Feature Flags
 
-- `default`: `wrappers`
-- `wrappers`: enable support for `atspi-common` wrapper types that categorize events by interface, as well as the all-encompassing `Event` enum that can store any event type.
-    - This also enables the `event_stream` function that allows you to receive a stream of `Event`s instead of specific events.
-- `tracing`: enable support for the `tracing` crate
+* `default`: `wrappers`, `p2p`
+* `p2p` : dependencies `async-executor",`dep:async-lock` and enables `zbus/p2p`
```
TTWNO, line 14: Remove async lock dep from comment.

TTWNO (approved):

PR #261, opened 2025-06-06
Add another tree example

C-Loftus:
I found the other atspi example code very useful and as such I wanted to contribute another example that is still relatively simple but helps to show how to:

- turn an `ObjectRef` into an `AccessibleProxy` and retrieve the natural language name
- get the coordinates by retrieving the associated `component` 
- checking the `StateSet` of the accessible

Some of this logic is shown in the existing tree example or Luuk's [a11y-app](https://github.com/luukvanderduim/a11y-app) but I tried to make it a bit less code / show off some other functionality,

## Example
```
host@computer ~/g/cAtspi (main)> cargo run --example doc-info-tree
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/examples/doc-info-tree`
  child, role:link, name:Skip to content, extents:(66, 116, 1, 1)
            child, role:push button, name:Open global navigation menu, extents:(85, 135, 38, 38)
      child, role:link, name:Homepage, extents:(138, 135, 38, 38)
                child, role:link, name:C-Loftus, extents:(186, 138, 79, 33)
                child, role:link, name:atspi, extents:(285, 138, 56, 33)
          child, role:push button, name:Type / to search, extents:(1578, 135, 312, 38)
      child, role:link, name:Chat with Copilot, extents:(1899, 135, 38, 38)
      child, role:push button, name:Open Copilot…, extents:(1938, 135, 28, 38)
      child, role:push button, name:Create something new, extents:(1996, 135, 69, 38)
    child, role:link, name:Your issues, extents:(2075, 135, 38, 38)
    child, role:link, name:Your pull requests, extents:(2123, 135, 38, 38)
      child, role:link, name:You have no unread notifications, extents:(2171, 135, 38, 38)
            child, role:push button, name:Open user navigation menu, extents:(2219, 135, 40, 38)
          child, role:link, name:Code, extents:(85, 194, 90, 36)
          child, role:link, name:Pull requests, extents:(184, 194, 157, 36)
          child, role:link, name:Actions, extents:(351, 194, 110, 36)
          child, role:link, name:Projects, extents:(471, 194, 115, 36)
          child, role:link, name:Wiki, extents:(596, 194, 85, 36)
          child, role:link, name:Security, extents:(690, 194, 116, 36)
          child, role:link, name:Insights, extents:(816, 194, 116, 36)
          child, role:link, name:Settings, extents:(942, 194, 117, 36)
[...]

PR #268, opened 2025-06-19
Add an example for text underneath the mouse

### LINKED ISSUE #263: (atspi-proxies): `PartialEq` on `AccessibleProxy` likely wrong
Consider two `AccessibleProxy`s, on different applications, but both referencing the root node path.
Then according to this definition these would be the same.

```Rust
impl PartialEq for AccessibleProxy<'_> {
    fn eq<'a>(&self, other: &Self) -> bool {
        (self.inner().path() == other.inner().path())
    }
}
```

C-Loftus:
- **Remove Eq, PartialEq for AccessibleProxy; Fix #263 (#264)**
- **Add a registry root method (#265)**
- **Succinct DFS traversal example with natural language app names (#261)**
- **StartUnderText**
- **Add an example for getting the text underneath the mouse**

### DIFF
--- a/atspi/Cargo.toml
+++ b/atspi/Cargo.toml
@@ -39,6 +39,11 @@ harness = false
 name    = "event_parsing_100k"
 path    = "./benches/event_parsing_100k.rs"
 
+
+[[example]]
+name              = "text-under-mouse"
+path              = "./examples/text-under-mouse.rs"
+
 [[example]]
 name              = "tree"
 path              = "./examples/bus-tree.rs"
--- a/atspi/examples/text-under-mouse.rs
+++ b/atspi/examples/text-under-mouse.rs
@@ -1,0 +1,78 @@
+//! This example demonstrates how to get the currently selected text.
+//!
+//! ```sh
+//! cargo run --example text-under-mouse
+//! ```
+//! Authors:
+//!    Colton Loftus
+
+use atspi::MouseEvents;
+use atspi_connection::set_session_accessibility;
+use atspi_proxies::{accessible::ObjectRefExt, proxy_ext::ProxyExt};
+use futures_lite::stream::StreamExt;
+use std::error::Error;
+
+#[async_std::main]
+async fn main() -> Result<(), Box<dyn Error>> {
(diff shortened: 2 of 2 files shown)


TTWNO:
If this works in X, I am fine merging it. Under the condition that there is a big waning in the top level docs for the example

TTWNO:
Ouch! Never got it to work in X even?

Sadly, I think we should close it; no sense committing code that doesn't work.

I don't want this to discourage you at all; we all have ideas that don't work out!

TTWNO:
> If I tinker more I will do it on my fork

And if you figure it out, then come back and PR it again!

C-Loftus:
Had a great conversation on matrix with others about this today; writing it up here to summarize so I don't forget / for future reference:

Mouse events are coming from the registry; by watching for the mouse event, you can get the x and y coords you will need to find the associated text; note with the accessible coming from this event it doesn't have the text interface and as such to get the associated text you need to query it on the relevant frame of the desired app. This essentially means that you need to  get the root of the registry and then iterate through all the apps until you find the one that was clicked on / focused. Generally the application itself does not implement the text interface, only its windows, hence why the above Firefox example was failing. 

Then you need to iterate through all its frames until you find the one where the user clicked. Because this is a lot of dbus round trip, Orca apparently uses `libwnck` to get this window metadata without needing to do the round trips. 

libwnck is used, because, to quote bgt: 

> well, think on this: how do you know in the bounds of which window you are? how do you know what frame you should call this on? so you either go through every frame under the root application and try that until it doesn't error, for every single mouse move event, which is costly as hell as you could imagine, or use wnck to find out what window has focus, where it is, what screen you're on perhaps, etc

As such this is dependent on X11

Once you have the root acessible of the desired frame of the desired app, you can then use the equivalent of `get_descendant_at_point` in libatspi to recursively find the most deep node in the accessibility tree that overlaps with the relevant x, y coordinate. We have to use this since multiple accessible's could be overlapping. For instance, if a user clicks on a website navbar, you need to descend into the navbar, then into a button component, and then finally get the text.

PR #269, opened 2025-06-20
Use root helper instead of exposing dbus destination

C-Loftus:
Change the example to use the root helper instead of exposing the raw dbus connection which may be less intuitive to a new user


luukvanderduim (approved):
Nice!

Issue #271, opened 2025-06-22
Handle empty application names


TTWNO:
Recently came across this with GTK4, they follow the specification [defined in Cache.xml:49](https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/main/xml/Cache.xml?ref_type=heads#L49) which allows the `s` in an `(so)` pair where `o` = `/org/a11y/atspi/accessible/null` to be empty.

However, our `ObjectRef` always expects an `OwnedUniqueName`, and there is no option for failure in decoding this.

Therefore, you will get a `Variant(Message("Invalid bus name. See https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus"))` when trying to decode the `(so)` pair.

The easiest way to reproduce this is using the `Cache::get_items` method. I've create a [simple example to reproduce the issue here](https://github.com/TTWNO/a11y-app/tree/cache-get-items-test) (branch `cache-get-items-test`) using Luuk's `a11y-app` as a starting point.

To run it, just start the `gtk4-demo-application`, then run the application with `cargo run -- gtk4-demo-application`.

TTWNO:
I think we need to reconsider the shape of `ObjectRef`. Either:

1. `name` becomes `Option<Uniquename<'_>>`
2. `name` becomes `String`

Or we consider making an enum.

```rust
enum ObjectRef {
  Null, // `/org/a11y/atspi/accessible/null`
  Root, // `/org/a11y/atspi/accessible/root`
  Item(UniqueName, ObjectRef),
}
```

cc @DataTriny @luukvanderduim 

As this will affect AccessKit too.

TTWNO:
Looks like this is a pretty common thing, I'll reference as I file issues upstream:

1. [GTK](https://gitlab.gnome.org/GNOME/gtk/-/issues/7617)

luukvanderduim:
> Don't quite have time to go through stuff as I'm just heading out, but wanted to drop my findings before I did.
> 
> GTK4 also produces inconsistent parent information:
> 
>     1. Start `gtk-demo-application`
> 
>     2. Run the following commands:
>        
>        * `busctl --user get-property --address unix:path=/run/user/1000/at-spi/bus_0 ':1.36' /org/a11y/atspi/accessible/root org.a11y.atspi.Accessible Parent`
[...]

PR #272, opened 2025-06-23
fix(common/object-ref): Translate "/org/a11y/atspi/null" sentinel value (issue #271)

luukvanderduim:
GTK4 may return empty strings as bus name, but `atspi` stores bus names more strictly typed.

In `atspi` the `ObjectRef`'s `Default` implementation is the null object.

This PR adjusts `ObjectRef`'s `Deserialize` implementation to return `ObjectRef::default()` if it encounters an `(so)` with an empty string and the null path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1900,9 +1900,9 @@ dependencies = [
 
 [[package]]
 name = "zbus-lockstep"
-version = "0.5.0"
+version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


luukvanderduim:
> Unfortunately, this will not work, as [the default is a valid bus address](https://github.com/odilia-app/atspi/blob/main/atspi-common/src/object_ref.rs#L24) which consumers may use to create connections to an application.

Our default has a bus name I have never seen given to a connection and the null path. Have you ever seen a connection that had ":0.0" as it's unique name?

So we can say the 'risk' already existed with our `ObjectRef::default()`

PR #277, opened 2025-07-15
fix(proxies): Method `nactions` remove tab in the commnets

luukvanderduim:
The tab makes it appear we start a code block and suddenly code is expected where we don't have any.

Because the tab was perfectly aligned with the rest, this gets the 2025 Obscured Error Award.


decathorpe:
Note that there still appear to be dozens of instances of stray `\t` characters across the whole project, especially in doc comments - they just appear inside the text so they don't matter as much. I wonder how this happened ...

PR #283, opened 2025-07-22
chore: update Cargo.lock with minimal direct versions

luukvanderduim:
update Cargo.lock with minimal direct versions


luukvanderduim:
We don't want to keep minimal versions in the lockfile.

PR #286, opened 2025-08-08
Add an example showing how to get the currently focused frame

C-Loftus:
The currently focused frame is useful for doing many types of traversals. For instance, if a user wants to find an accessible at a given coordinate, they would need to use the currently focused frame. The top level accessible from the registry that represents the app does not output this info. Rather, you need to get it from the child frames. I presume this is the case for many other types of visual / user-facing info so it seems like calling out how to get the frames would be useful.

I was using this to try and implement #285 but even though I am still stuck on that, it seems useful to add this example for posterity.

In this example I put a few assertions that help new users understand the properties of the example.

### DIFF
--- a/atspi/Cargo.toml
+++ b/atspi/Cargo.toml
@@ -41,6 +41,10 @@ name    = "event_parsing_100k"
 path    = "./benches/event_parsing_100k.rs"
 
 [[example]]
+name = "currently-focused-frame"
+path = "./examples/currently-focused-frame.rs"
+
+[[example]]
 name = "accessible-counts"
 path = "./examples/accessible-counts.rs"
 
--- a/atspi/examples/currently-focused-frame.rs
+++ b/atspi/examples/currently-focused-frame.rs
@@ -1,0 +1,46 @@
+//! This example prints out the currently focused frame. A
+//! frame is generally semantically equivalent to an application
+//! window.
+//!
+//! ```sh
+//! cargo run --example currently-focused-frame
+//! ```
+//! Authors:
+//!    Colton Loftus
+
+use std::error::Error;
+
+use atspi::{MouseEvents, State};
+use atspi_connection::set_session_accessibility;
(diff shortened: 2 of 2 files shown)


C-Loftus:
Believe the clippy error is coming from code outside this PR

TTWNO:
If you rebase on main it should go away :)

TTWNO:
Since [`async-std` is deprecated](https://github.com/async-rs/async-std/blob/main/README.md), I don't want to add any more examples using it. They were using `smol` under the hood for some time prior to the deprecation so it really just a swap out.

Otherwise, LGTM!

TTWNO (approved):

TTWNO:
> Why do you need to query for every application state? It's paying a D-Bus call for nothing since it is not relevant to the task you are trying to accomplish.

To find the [active (currently focused) application](https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/main/xml/Accessible.xml?ref_type=heads#L774)? How would you find it?

C-Loftus:
> Why do you need to query for every application state? It's paying a D-Bus call for nothing since it is not relevant to the task you are trying to accomplish.

Assuming you mean the code below, I am getting the state just so I can call out to the user one of the properties of frames and their relation to the top level app. Namely that in this case that the top level app shouldn't have active state. In my opinion, I feel that adding extra info like this that calls out semantics of atspi that aren't explicitly stated in the client API are useful and help new users more than optimizing for dbus calls. Let me know if I am misinterpreting though or, as tait says, if there is a better way.

```rs
		let state = proxy.get_state().await?;
		assert!(!state.contains(State::Active), "The top level application should never have active state; only its associated frames should have this state");
```

TTWNO:
Is there a distinction between window and application in the documentation?

TTWNO:
Finals are over; let's get all your PRs merged!

PR #288, opened 2025-08-10
Add an example for getting text under the mouse click

C-Loftus:
This example shows how you can click on text and print it out to stdout.

- It reads the mouse events in a separate tokio task and then processes it in the main one.
- It converts mouse events from absolute screen coordinates to window relative ones so that they work with `get_accessible_at_point`.
- It iteratively descends down the tree and returns the component with text once it is found
    - Note, in my testing most of the time it doesn't actually need to descend for any reason since `get_accessible_at_point` gets the proper accessible immediately; however, Orca does it this way along with checking the state while descending, so I am inclined to keep it as is
- Once it gets the final relevant accessible, it gets the offset and uses the offset to get all the relevant text
    -  Note, I was previously using `i32::MAX` to get all text however this always returns the empty string literal `''`. It seems that if you want to get all text you must use `-1` which was unclear to me. It is probably best either way to do it as I do now, namely explicitly getting the text range and then using that range to get the string, instead of giving an arbitrary value like -1. 

## Limitations 

- You need to make sure you are clicking on the proper location. i.e. in Firefox sometimes you may want to click slightly ahead of the text you want to have output. It seems like this coord behavior may change depending on app
- This example works on X11 but not Wayland due to the lack of global coords on Wayland 
- This seems to work in Firefox but in other apps often give errors like
```
Error: Variant(Message("Invalid unique name. See https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus"))
```
- Certain pages even in Firefox don't work for some reason. I suspect this is due to the fact that the underlying HTML semantics may be wrong. However, this is quite hard to debug and perhaps out of scope to make sure all edge cases work. 

## Background

This took me a long time to figure out but with huge thanks to @DataTriny, I was able to figure it out and resolve the discussion in #285

### DIFF
--- a/atspi/examples/text-under-click.rs
+++ b/atspi/examples/text-under-click.rs
@@ -1,6 +1,6 @@
 //! This example prints out the text under the mouse click.
 //! This example only works with X11 given the fact
-//! Wayland does not support global coordinates or 
+//! Wayland does not support global coordinates or
 //! global input events.
 //!
 //! ```sh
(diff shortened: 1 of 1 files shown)


C-Loftus:
Thanks very much for the feedback Luuk! I will be a bit busy the next few days but can definitely add PRs for your comments in the rustdocs by this Friday. A few initial comments:

>  Why does it require the mouse click do you think?

I was just using mouse click since there is a risk of many dbus calls and thus I assumed that text under the mouse hover location had too much throughput to be useful in production.

> Just speculating, but you rely on click. What if UI elements just are not focusable?

Yup that definitely could be an issue. In fact if I recall, I was having issues on Wikipedia; now that I think about it, I believe on Wiki pages the main paragraph content isn't focusable. Perhaps that is it. However, I will say that I am just using the click to get the location of the click, so I thought that would allow for getting the text of non-focusable elements.

PR #289, opened 2025-08-11
Make MatchType::Invalid default variant

TTWNO:
Previously, the default `MatchType` variant was `MatchType::All`.

However, this does not interact nicely with the `ObjectMatchRuleBuilder` since this means even on properties which are not specified, the specified match type would still be `MatchType::All`.

What this meant in Odilia is that it would perform an "All" match on every preperty (which can sometimes cause an await point) even when the set of things to check was empty.

This fixes that issue, since the unspecified properties now get set to `MatchType::Invalid` (i.e.: do not perform search of this property). This has been updated in the documentation as well.


luukvanderduim (approved):
Well spotted, thanks!

PR #294, opened 2025-08-17
`ObjectRef` refactor to address #271, supersedes #272

luukvanderduim:
Problem: 
AT-SPI2 allows implementations to return an empty string as bus name to indicate the absence of a bus name.
However, `atspi` stores bus names more strictly typed.

This addresses #271 and supersedes #272 

This PR refactors `ObjectRef`

```rust
/// A unique identifier for an object in the accessibility tree.
///
/// A ubiquitous type used to refer to an object in the accessibility tree.
///
/// In AT-SPI2, objects in the applications' UI object tree are uniquely identified
/// using a applications' bus name and object path. "(so)"
///
/// Emitted by `RemoveAccessible` and `Available`
#[validate(signal: "Available")]
#[derive(Clone, Debug, Eq, Type)]
#[zvariant(signature = "(so)")]
pub enum ObjectRef<'o> {
	Null,
	Owned { name: UniqueName<'static>, path: ObjectPath<'static> },
	Borrowed { name: UniqueName<'o>, path: ObjectPath<'o> },
} 
```
Due to `ObjectRef`'s central role in the accessibility tree, this change was pervasive and had consequences all over the tree.


atspi/examples/currently-focused-frame.rs
```diff
@@ -26,11 +26,19 @@ async fn main() -> Result<(), Box<dyn Error>> {
 	let mut found_active_frame: bool = false;
 
 	for app in apps.iter() {
+		if app.is_null() {
```
TTWNO, line 29: Similarly here. From the outside (of atspi land) this looks like nonsense. I can see the confusion now: "Why would the application registry send back a null reference?"

TTWNO (approved):

PR #314, opened 2025-09-18
`GetNlinks/selection` -> `GetNLinks/Selection`

TTWNO:
- Searched through XML docs using the following: `rg 'method name="[A-Za-z]+"' atspi-common/xml/ -o | rg '[A-Z]{2,}'`
- Found exactly four matches:
    - `name="GetMDIZOrder"`
    - `name="GetNSelections"`
    - `name="GetNLinks"`
    - `name="GetURI"`
- All now use their correct name via either `_` separation or `#[zbus(name = )]`


luukvanderduim:
It appears `zbus_xmlgen` added annotation with attributes if names did not correspond with the generated member name not too long after we last generated our proxies.

Just to be sure, I decided to regenerate proxy definitions from XML and look for all changed name attributes and see whether we were indeed complete.
I also decided to make all those instances explicit (by decorating the items) regardless of whether we had fixed them with method name change - just  to make the change and member name visible.

Issue #327, opened 2026-05-27
Bill of Materials is unnecessarily long


luukvanderduim:
# Bill of Materials is unnecessarily long

Our `Cargo.lock` currently includes dependencies for Windows and macOS. This makes our dependency list (BOM) longer than it needs to be.

Because of this, audit tools (like `cargo-audit`) might complain about vulnerabilities in Windows crates that we don't even use. It also forces Cargo to download and index metadata we don't need.
Additionally, binaries and libraries that have `atspi` as a dependency inherit our dependency list.

Luckily we can improve upon our BoM!

By adding `filter-platform` to `.cargo/config.toml`, we can limit the lockfile to just Linux, BSD, and WASM. Testing this showed it instantly drops 25 crates from the lockfile.

luukvanderduim:
The solution mentioned was fabricated by our automated stochastic pattern recombinating friend and I fell for it.

[This closes the issue / PR](https://github.com/odilia-app/atspi/pull/328#issuecomment-4562915302)

PR #328, opened 2026-05-27
Reduce Bill of Materials (#327)

luukvanderduim:
# Clean up Cargo.lock

Our `Cargo.lock` currently includes dependencies for Windows and macOS. This makes our dependency list (BOM) longer than it needs to be.

Because of this, audit tools (like `cargo-audit`) might complain about vulnerabilities in Windows crates that we don't even use. It also forces Cargo to download and index metadata we don't need.

~~By adding `filter-platform` to `.cargo/config.toml`, we can limit the lockfile to just Linux, BSD, and WASM. Testing this showed it instantly drops 25 crates from the lockfile.~~

**edit** The above was confabulated by Gemini 3.5. It had mixed up a `cargo metadata` / `cargo tree` flag `--filter-platform`

Results below were likely coincidental due to generating a new Cargo.lock.

**edit2** There may be a way to do this in CI - I'll look to implement this instead.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,19 @@
+# This [net] section filters the Cargo.lock to reduce the "Bill of Materials" (BOM)
+# by excluding dependencies for unsupported platforms like Windows and macOS.
+
+[net]
+filter-platform = [
+    # Linux & Android targets
+    "x86_64-unknown-linux-gnu",
+    "aarch64-unknown-linux-gnu",
+    "i686-unknown-linux-gnu",
+    "armv7-unknown-linux-gnueabihf",
+
+    # BSD variants
+    "x86_64-unknown-freebsd",
+    "x86_64-unknown-netbsd",
+    "x86_64-unknown-openbsd",
+
+    # WebAssembly (for atspi-common)
+    "wasm32-unknown-unknown"
+]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,25 +3,10 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
-dependencies = [
- "gimli",
-]
-
-[[package]]
-name = "adler2"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
-
-[[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
 ]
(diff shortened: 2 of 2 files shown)


luukvanderduim:
During cargo-msrv it was found that zbus-macros requires edition 2024.

Either require a lower zbus version, limiting upstream users, or 
migrate to edition = "2024" forcing some upstream users to follow suit. 
Edition 2024 was introduced in 1.85.0, current stable is 1.95.0

```bash
Run cargo test --workspace --no-run --all-features
warning: unused config key `net.filter-platform` in `/home/runner/work/atspi/atspi/.cargo/config.toml`
    Updating crates.io index
 Downloading crates ...
  Downloaded async-io v2.6.0
  Downloaded anes v0.1.6
  Downloaded static_assertions v1.1.0
  Downloaded zbus_macros v5.15.0
error: failed to download replaced source registry `crates-io`

Caused by:
  failed to parse manifest at `/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zbus_macros-5.15.0/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.77.2 (e52e36006 2024-03-26)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
Error: Process completed with exit code 101.

```

TTWNO:
I'm happy to migrate to the 2024 edition.

luukvanderduim:
> I'm happy to migrate to the 2024 edition.

Great! I'll bump MSRV / editions (I suppose both are required) and let's see what CI thinks then.

luukvanderduim:
So this is painful - and embarrassing.

I needed to adjust the above and need to rethink this PR because the solution as provided was hallucinated. The results were a coincidence.

~~By adding `filter-platform` to `.cargo/config.toml`, we can limit the lockfile to just Linux, BSD, and WASM. Testing this showed it instantly drops 25 crates from the lockfile.~~

**edit** The above was confabulated by Gemini 3.5. It had mixed up a `cargo metadata` / `cargo tree` flag `--filter-platform`

Results were coincidental due to generating a new Cargo.lock which included zbus 5.15 which now uses edition 2024 -> resolver = 3, which means that it's dev-dependencies and dependencies are no longer unified, shrinking its dependency list.

TTWNO:
Good find, haha! I see how that could be confusing.

Thank you for clarifying and following up on AI-generated code. This is the responsible way to use it :)

luukvanderduim:
Honestly I am disgusted that it managed to convince me. That Google Clanker keeps lying like it's a proverbial 1980s car salesman?

TTWNO:
> That Google Clanker keeps lying like it's a proverbial 1980s car salesman?

It is a sneaky one indeed! Can we train an AI that checks AI facts? LOL

luukvanderduim:
We can move to edition 2024 and/or upgrade our zbus / z-* deps but that is no longer in line or spirit of this PR.
There is no way to trim Cargo.lock by eliminating target triples we know are eliminated at build time, and this is by design:

Having a feature in cargo came up several times but the cargo team say it is too complicated too realize while also keeping the promise ["that changing the version of cargo should not change your Cargo.lock."](https://github.com/rust-lang/cargo/issues/15563#issuecomment-2895511891) and [The cargo team's blog post discusses the matter](https://blog.rust-lang.org/inside-rust/2025/02/27/this-development-cycle-in-cargo-1.86/#specifying-supported-platforms-in-packages). From what I understand, "it's complicated" ..

The only thing we may consider is tell our audit-tool `cargo-deny` which target triples we want it to consider so it will never complain about a vulnerability in a transitive dependency for Windows or MacOS for instance.

TTWNO:
> The only thing we may consider is tell our audit-tool cargo-deny which target triples we want it to consider so it will never complain about a vulnerability in a transitive dependency for Windows or MacOS for instance.

Now that is a good idea!!

PR #329, opened 2026-05-28
feat(proxy): add find_by_name to RegistryProxy (closes #251)

### LINKED ISSUE #251: Add `RegistryProxy` method to get root `ObjectRef` by application name
In [a11y-app](https://github.com/luukvanderduim/a11y-app) I've written code to retrieve a root object by looking it up by name.

There may be more than one connection (bus-names) associated with the same application name. This means that the return type should probably be a `Vec<ObjectRef>`, also, there may be choices to be had to how lenient the function should match on names.

The following comes to mind:
`find_by_name(&self, name: &str)` for exact matching and `find_by_name_with<F: FnOnce>(&self, F)` for user defined matches?

JustinCMR:
Closes #251.

Adds two helpers on `RegistryProxy` for locating application roots by accessible `Name`:

```rust
pub async fn find_by_name(&self, name: &str)
    -> Result<Vec<ObjectRefOwned>, AtspiError>;

pub async fn find_by_name_with<F>(&self, predicate: F)
    -> Result<Vec<ObjectRefOwned>, AtspiError>
where
    F: FnMut(&str) -> bool;
```

`find_by_name` does exact matching; `find_by_name_with` accepts a caller-defined predicate (case-insensitive, prefix, regex).

## Notes

- Return type is `Vec<ObjectRefOwned>` to match `AccessibleProxy::get_children`'s existing idiom; multiple matches are possible since more than one connection may publish a root with the same `Name`.
- Predicate is `FnMut(&str) -> bool` rather than the issue's pseudocode `FnOnce`, since it is invoked once per child.
- Per-child failures (process gone, denied read, transient DBus error) are skipped rather than aborting the walk.

## Verified

- [x] `cargo build --workspace`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test --workspace --lib` (1142 passed)
- [x] `cargo doc --no-deps -p atspi-proxies`

Integration smoke deferred (no live AT-SPI session from Windows).

### DIFF
--- a/atspi-proxies/src/registry.rs
+++ b/atspi-proxies/src/registry.rs
@@ -155,7 +155,11 @@
 //! [ac_de]: ../../atspi-connection/struct.AccessibilityConnection.html#method.deregister_event
 //! [ac]: ../../atspi-connection/struct.AccessibilityConnection.html
 
+use atspi_common::object_ref::ObjectRefOwned;
 use zbus::names::OwnedBusName;
+
+use crate::accessible::{AccessibleProxy, ObjectRefExt};
+use crate::AtspiError;
 
 // The proxy macro attribute `assume_defaults = false` to avoid generating defaults service and path
 // The generated defaults don't make sense in AT-SPI2 / accessibility-bus context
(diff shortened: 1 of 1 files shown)


luukvanderduim:
Thanks for the PR, much appreciated. 
Code looks great as well. Good stuff!

To get CI running and subsequent review we really require contributors to sign their commits. If you could do so that would be fantastic. 
Thanks in advance!

JustinCMR:
Done. Signed the commit and rebased onto `main` (resolved a small import conflict in `registry.rs` from the recent module-doc rework). Ready for CI.

atspi-proxies/src/registry.rs
```diff
@@ -178,3 +182,84 @@ pub trait Registry {
...
+		// `AccessibilityConnection::root_accessible_on_registry`.
+		let root_accessible = AccessibleProxy::builder(conn)
+			.destination(inner.destination().clone())?
+			.path(atspi_common::ACCESSIBLE_ROOT_PATH)?
+			.cache_properties(zbus::proxy::CacheProperties::No)
+			.build()
+			.await?;
+
+		let children = root_accessible.get_children().await?;
+
```
luukvanderduim, line 246: Could we use [Vec::retain](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.retain) here?

atspi-proxies/src/registry.rs
```diff
@@ -178,3 +182,84 @@ pub trait Registry {
...
+	///
+	/// [`find_by_name_with`]: RegistryProxy::find_by_name_with
+	///
+	/// # Errors
+	///
+	/// Errors if the registry view cannot be built or [`get_children`] fails.
+	///
+	/// [`get_children`]: crate::accessible::AccessibleProxy#method.get_children
+	pub async fn find_by_name(&self, name: &str) -> Result<Vec<ObjectRefOwned>, AtspiError> {
+		self.find_by_name_with(|candidate| candidate == name).await
```
luukvanderduim, line 206: I love the delegation!

luukvanderduim (changes requested):
I really like the PR - thank you. And thanks for signing.

I realize compilable / runnable doctests are difficult in this context because we are looking for a child of `Registry` which there is none always around.

I think I spotted one nice opportunity for improvement, which is to use `Vec::retain` on `children` and mutate it in-place!

atspi-proxies/src/registry.rs
JustinCMR: Thanks! I kept this as an explicit loop as the match is on the accessible Name property (#[zbus(property)]), a per-child async read (as_accessible_proxy(…).await, then .name().await) before the predicate can run. Vec::retain's closure is synchronous,  it cant await the name lookup. Thats the same reason from_app_name in a11y-app collects into a separate Vec rather than retaining in place. Happy to reshape if you'd prefer, but this keeps it consistent and dependency free.

TTWNO: Luuk is suggesting that you fully populate the `Vec`, then retain. We are not suggesting you put the entire async section of the loop within the retain filter.

luukvanderduim: Just for laughs. Obviously we cannot do what I had thought was smart.
`AccessibleProxy::get_children().await?` does not yield names, but children.

Wishful thinking on my part. Sorry!

luukvanderduim (approved):

luukvanderduim:
Thanks @JustinCMR !

PR #346, opened 2026-06-02
fix(conn): eliminate `block_on` from async fn

luukvanderduim:
`block_on` blocks the current thread in async context. Theoretically, this can hamper making progress on other tasks on the same executor.

Even though this call was performed before we could ever listen to a stream for instance (which requires continuous polling to prevent buffer-overruns) this does eliminate a minor footgun and eliminates the need for the futures-lite executor in our library.


TTWNO:
LGTM!

Issue #347, opened 2026-06-04
[proxies] `assume-defaults` should not be set `true`


luukvanderduim:
We have had a bit of a misunderstanding about configuring defaults in atspi-proxies.

The assume-defaults = bool key-value pair in the `zbus::prtoxy` macro, if set to `true`, will assume the default_service equals interface which in the context of AT-SPI is almost always false because the bus
has many services that (for instance) implement `Accessible`. 

Most of our builder-helpers probably override the default - but - as I found out, using the `Proxies` helper one can run into falsely assumed defaults. The fix is easy, just don't assume defaults for any (barring the exception perhaps) interface. The per-interface documentation requires knowing whether the particular interface is implemented always, sometimes for an application and once or for each node of an application.

luukvanderduim:
This is why we cannot `assume_defaults`:
[snip from proxy proc-macro in zbus-macros](https://docs.rs/crate/zbus_macros/latest/source/src/proxy.rs#197-203)

luukvanderduim:
Fixed in #348

PR #348, opened 2026-06-04
fix(proxies): fix `assume_defaults` on interfaces (issue: #347)

luukvanderduim:
The `zbus::proxy` macro allows users to assume
defaults for `path` and `service` parameters.
This does make sense on the session bus where
there's often only a single implementer of a
given interface, with a fixed service and path,
but things are different on the a11y bus.

Addresses issue: #347


luukvanderduim:
@TTWNO I have no idea actually what the DeviceEvent interfaces do or when one would use these.

I tried to write up some basics as I went along but nothing is perfect. Feel encouraged to clone and improve where ever you can. Thanks in advance.

luukvanderduim:
We want to review all builders in the whole source tree whether they were relying on bad derived defaults for either path or destination (by omitting the method on the builder).

Accessible has a sane default for root, 
Application, similar.
Registry, Socket, Status and Bus have sane defaults for all parameters.
Cache has a default path (which seems correct)
Collection's default has a sane default path.

Otherwise only interfaces are configured and no defaults derived.

luukvanderduim:
> We want to review all builders in the whole source tree whether they were relying on bad derived defaults for either path or destination (by omitting the method on the builder).
> 
> Accessible has a sane default for root, Application, similar. Registry, Socket, Status and Bus have sane defaults for all parameters. Cache has a default path (which seems correct) Collection's default has a sane default path.
> 
> Otherwise only interfaces are configured and no defaults derived.

Reviewed builders.

DataTriny:
- The `Collection` interface can be advertised by any object and I have no idea where your default path come from.
- Setting the root path as the default for the `Accessible` proxy doesn't seem right given that most objects don't live there.
- The `Socket` interface can be implemented on any object that acts as the socket for the plug of an out-of-process tree. This is not a common use case so the default path is probably fine, the documentation is not though.

luukvanderduim:
> * The `Collection` interface can be advertised by any object and I have no idea where your default path come from.

It was here, can't remember where it came from either, happy you chime in with these corrections! Thanks!
> 
>     * Setting the root path as the default for the `Accessible` proxy doesn't seem right given that most objects don't live there.

We think it would be convenient, but it is certainly not a hill I am willing to die on.
Will change accordingly! Thanks again.

> 
>     * The `Socket` interface can be implemented on any object that acts as the socket for the plug of an out-of-process tree. This is not a common use case so the default path is probably fine, the documentation is not though.

Could you elaborate on when would one have or want an out of process tree? Is that a (part of) a UI tree that is not exposed on the a11y bus?

Thanks Arnold, much appreciated!

luukvanderduim:
> * The `Collection` interface can be advertised by any object and I have no idea where your default path come from.

I remember we recognized it as a pattern, but I don't think we saw it somewhere written down. To me it made sense that providers would expose one object that knew of the whole application's tree. @DataTriny Why is it deemed useful to implement it for arbitrary nodes instead of a central node?

DataTriny:
The socket/plug mechanism is used to link two AT-SPI trees that are managed by distinct processes. One popular example would be GNOME Web and WebKitGTK: the browser shell and the web view are isolated in their own process yet they appear as one tree to assistive technologies.

DataTriny:
> Why is it deemed useful to implement it for arbitrary nodes instead of a central node?

The object on which Collection methods are called serve as the point of reference for iterating/filtering siblings/descendants... Anyway the whole API is bad.

luukvanderduim:
> The socket/plug mechanism is used to link two AT-SPI trees that are managed by distinct processes. One popular example would be GNOME Web and WebKitGTK: the browser shell and the web view are isolated in their own process yet they appear as one tree to assistive technologies.

Thank you, I have read the text we had on plug several times but it did not make sense, but this example does. Perfect.

luukvanderduim:
> > Why is it deemed useful to implement it for arbitrary nodes instead of a central node?
> 
> The object on which Collection methods are called serve as the point of reference for iterating/filtering siblings/descendants... Anyway the whole API is bad.

Sure, the API has it's shortcomings..

luukvanderduim:
@DataTriny If you find any more falsehoods and/or mistakes, they'll find their way on my todo ASAP. I want these fixes and corrections in.

atspi-proxies/src/accessible.rs
```diff
@@ -21,7 +80,7 @@ use zbus::names::BusName;
 #[zbus::proxy(
 	interface = "org.a11y.atspi.Accessible",
 	default_path = "/org/a11y/atspi/accessible/root",
```
TTWNO: Looking at this closer, I think the default should be removed.

Provide a static root (which wr already do) and call it a day. Leave a doc comment explaining how to use it.

atspi-proxies/src/action.rs
```diff
@@ -1,21 +1,56 @@
...
+//!
+//! ## How to obtain an `ActionProxy`
+//!
+//! There are two idiomatic ways to obtain an `ActionProxy`:
+//!
+//! ### 1. Safe conversion via [`ProxyExt`][pe] (Recommended)
+//! If you already have an [`AccessibleProxy`][ap] for an actionable node, you can
+//! safely query and convert it using the [`ProxyExt`][pe] trait:
+//!
+//! ```rust,ignore
```
TTWNO: Use no_run insread of ingore. Ignoring means this will not even be compiled. Same for all new doc tests with the same pattern.

TTWNO:
Review in progress. Will finish this afternoon.

DataTriny:
@luukvanderduim The documentation for `Socket` is wrong: `Embed` only exists on the registry, `Embedded` is the entrypoint for the socket/plug dance, and in this context the Socket interface is implemented on the remote object, not the host application.

luukvanderduim:
> @luukvanderduim The documentation for `Socket` is wrong: `Embed` only exists on the registry, `Embedded` is the entrypoint for the socket/plug dance, and in this context the Socket interface is implemented on the remote object, not the host application.

Got it - on it.

luukvanderduim:
> > @luukvanderduim The documentation for `Socket` is wrong: `Embed` only exists on the registry, `Embedded` is the entrypoint for the socket/plug dance, and in this context the Socket interface is implemented on the remote object, not the host application.
> 
> Got it - on it.

So the dance order is: 

The socket application calls Socket::embedded
    Registry informs the Plug application
Plug applcation sets socket application's socket object as parent.
Plug application calls `Socket::embed` on the registry to identify itself.
    Registry sets the "Id" property on the `org.a11y.atspi.Application` interface on the plug object.
    The Embed method returns with the bus name and object path for the registry's root object.
    
(It says "Registry's root path and object" but I can see how that should "registree" as in the one to register with?

Thanks in advance for clarifying.

luukvanderduim:
All interface proxy modules are now revised (again) and hopefully I got a bit more right this time.

atspi-proxies/src/accessible.rs
```diff
@@ -3,26 +3,87 @@
...
+//! [tp]: crate::text::TextProxy
+//! [cp]: crate::component::ComponentProxy
+//! [vp]: crate::value::ValueProxy
 
 use crate::common::{InterfaceSet, ObjectRef, RelationType, Role, StateSet};
 use crate::AtspiError;
 use atspi_common::object_ref::ObjectRefOwned;
 use zbus::names::BusName;
 
+pub const ACCESSIBLE_ROOT_PATH: &str = "/org/a11y/atspi/accessible/root";
```
TTWNO: I am 99% sure we have a root defined in `atzpi-common` that should be used here.

atspi-proxies/src/device_event_controller.rs
```diff
@@ -1,14 +1,46 @@
...
-//! You may prefer to adapt it, instead of using it verbatim.
+//! Historically, the `DeviceEventController` (DEC) interface, hosted by the central
+//! AT-SPI registry daemon, allowed assistive technologies (like screen readers) to
+//! intercept global hardware input events (keystrokes, mouse buttons, pointer moves)
+//! and synthesize input events (keyboard and mouse macros).
 //!
-//! More information can be found in the
-//! [Writing a client proxy](https://dbus.pages.freedesktop.org/zbus/client.html)
-//! section of the zbus documentation.
+//! ## ⚠️ Deprecation & Wayland Warning (Legacy Only)
```
TTWNO: If we are going to have such a strong deprecation warning, feature flag the whole file. Something like "legacy_x11_key_handling". Or, at minimum, add the `#[deprecated]` attribute.

atspi-proxies/src/device_event_listener.rs
```diff
@@ -1,18 +1,38 @@
...
-//! You may prefer to adapt it, instead of using it verbatim.
+//! The `DeviceEventListener` is the interface implemented **by the AT client itself**
+//! (such as a screen reader) and registered with the `DeviceEventController`. When a
+//! registered key or mouse event occurs, the registry calls [`notify_event`] on the
+//! client's listener proxy.
 //!
-//! More information can be found in the
-//! [Writing a client proxy](https://dbus.pages.freedesktop.org/zbus/client.html)
-//! section of the zbus documentation.
+//! ## ⚠️ Legacy Status Warning
```
TTWNO, line 11: Second of the previous comment. Mark as deprecated, not just mention it in the docs.

atspi-proxies/src/registry.rs
```diff
@@ -1,21 +1,94 @@
...
 //!
+//! * **Register for Events**: Subscribe globally to specific event streams (such as
+//!   `object:state-changed:focused` or `document:load-complete`) using [`register_event`].
+//! * **Deregister**: Unsubscribe when those events are no longer needed using [`deregister_event`].
+//! * **Query**: Inspect which clients have subscribed to which events via [`registered_events`].
+//!
+//! Once registered, the central registry daemon will ensure that whenever an application
+//! triggers the matching event, a D-Bus signal is routed to your client.
+//!
+//! ## 💡 Why register events here? (Why not just use D-Bus Match Rules?)
```
TTWNO: The emojis are not necessary. I don't mind LLM-help generating the docs, but that's just too much LOL!

atspi-proxies/src/collection.rs
```diff
@@ -1,24 +1,96 @@
...
+//! ).await?;
+//!
+//! for node_ref in matches {
+//!     println!("Found matching actionable node: {node_ref:?}");
+//! }
+//! ```
+//!
+//! [pe]: crate::proxy_ext::ProxyExt
+//! [ap]: crate::accessible::AccessibleProxy
+//! [mrb]: atspi-common::object_match::ObjectMatchRuleBuilder
```
TTWNO: This proxy's docs are incomplete unless it mentions the many cases where the interface is available, but does not work. The best example (citation needed) is Chromium. Which always returns an empty list no matter what the query is composed of.

Can you please confirm this? Adapt one of our examples that crawls a tree, and when a note implements the collection interface try looking for something that ought to exist (like _any child_), then compare with the children list. LLM might be able to generate this, even.

atspi-connection/src/p2p.rs
```diff
@@ -772,17 +783,23 @@ impl P2P for crate::AccessibilityConnection {
 			})
 			.cloned();
 
+		// This method relies on the default root path of AccessibleProxy
 		if let Some(peer) = lookup {
 			// If a peer is found, create an AccessibleProxy with a P2P connection
 			AccessibleProxy::builder(peer.connection())
+				.path(ACCESSIBLE_ROOT_PATH)?
```
TTWNO, line 790: Really wonder whether a `.root_path()` method may help us here. This will be a pretty commmon thing to write out.

atspi-connection/src/p2p.rs
```diff
@@ -772,17 +783,23 @@ impl P2P for crate::AccessibilityConnection {
...
+				.destination(peer.unique_name().clone())?
```
TTWNO, line 791: Likewise here, having an alternate method that can clone/borrow as needed could be beneficial to end-users as an extension trait.

If you agree, just make an issue for it and we can solve it in another PR.

atspi-proxies/src/application.rs
```diff
@@ -1,66 +1,80 @@
...
-//! It provides information about the application itself.
+//! ## Status and Deprecations
 //!
-//! ## Status
+//! Several methods and properties of this interface are legacy or considered
+//! deprecated, and should be avoided in modern implementations:
 //!
-//! A number of methods and properties of this interface have fallen in disuse or
-//! are / may be deprecated in the future.
+//! * [`id`](ApplicationProxy#method.id) & [`set_id`](ApplicationProxy#method.set_id)
```
TTWNO, line 15: Mark these methods deprecated, not just in docs.

atspi-proxies/src/socket.rs
```diff
@@ -1,21 +1,68 @@
...
+//!   web process.
+//!
+//! To assistive technologies (like screen readers), these must appear as a single,
+//! contiguous UI-tree.
+//!
+//! The entry point for the Plug/Embed mechanism is the `Embedded` method.
+//!
+//! 1. The `Embed` method is called by a socket application to inform the plug
+//!    application that it is being embedded.
+//!
```
TTWNO, line 25: This might appear as two lists, just remove the empty line.

atspi-proxies/src/socket.rs
```diff
@@ -1,21 +1,68 @@
...
+//! 2. The registry imforms the plug application of the socket object.
+//! 3. The plug application sets the socket object as its parent.
+//! 4. The plug application calls `Socket::Embed` to identify itself.
+//! 5. The Registry sets the `Id` property on the `org.a11y.atspi.Application`
+//!    interface on the plug object.
+//! 6. The method `Socket::Embed` returns with name and object path of the
+//!    Registry's root object.
```
TTWNO: I appreciate the list of events and the order they happen in, but for most other complex proxies there are little examples showing how to use this from the client perspective.

For this one, add an example with both client and server perspectives; should be just short enough that it's clear how this works from the code perspective. This is not exactly intuitive on its own.

TTWNO (changes requested):
Good work, Luuk! You continue to dive into areas of the code I haven't thought about in.. years, LOL! And you always fined something we/I did wrong a long time ago that could be improved. Thank you!

Just a few crossing Ts here and there and this should be good to go!

atspi-proxies/src/accessible.rs
luukvanderduim: Thanks! I removed that yesterday - or at least intended to but apparently it did not make it across the wire.

luukvanderduim: It says at the top this is outdated..

atspi-proxies/src/action.rs
luukvanderduim: I really don't like getting doctests to work. It is always a too long back-and-forth to get rustc satisfied. I get that we do it to avoid drift, the code (builder, proxies) have been tested extensively.

luukvanderduim: Luckily I only have to fix it once. All examples over all mod docs are the same.

atspi-proxies/src/application.rs
luukvanderduim: Good idea!

atspi-proxies/doc/socket-plug.md
```diff
@@ -0,0 +1,76 @@
+# Socket/Plug Stitching Mechanism
```
TTWNO: Include this in a separate section of the `socket.rs` file instead of being completely separate. You can even use the `#[doc = include_str!("...")]` macro to expand it in-place.

Then at least your Markup gets rendered.

atspi-proxies/doc/socket-plug.svg
TTWNO, line 1: Please include the source for generating this SVG (and note the tooling used to produce it). All images must have an alternative format for screen readers: if the source is relatively text-based, then that would be considered sufficient.

atspi-proxies/src/socket.rs
```diff
@@ -1,48 +1,128 @@
...
+//! so that an assistive technology sees a single, contiguous tree. This is purely a
+//! server-side orchestration; assistive technologies only consume the stitched result
+//! and do not take part in it.
+//!
+//! As far as we can tell from the reference implementation, the embedding itself happens
+//! directly between the host and the plug; the registry is only involved in the separate,
+//! general registration step. For a step-by-step walkthrough with a sequence diagram and
+//! source references, see the [Socket/Plug Stitching Mechanism] document.
+//!
+//! [Socket/Plug Stitching Mechanism]: https://github.com/odilia-app/atspi/blob/main/atspi-proxies/doc/socket-plug.md
```
TTWNO, line 26: 404 on the link. See earlier review comment about just including the doc in its own section.

atspi-proxies/src/registry.rs
```diff
@@ -1,21 +1,171 @@
...
+//!     .await?;
+//! # Ok::<(), atspi_common::AtspiError>(())
+//! # });
+//! ```
+//!
+//! [`register_event`]: RegistryProxy#method.register_event
+//! [`deregister_event`]: RegistryProxy#method.deregister_event
+//! [`registered_events`]: RegistryProxy#method.registered_events
+//! [or]: crate::common::ObjectRef
+//! [deref]: https://doc.rust-lang.org/std/ops/trait.Deref.html
```
TTWNO: Why use https?? Just use `std::ops::Deref`.

atspi-proxies/src/registry.rs
```diff
@@ -1,21 +1,171 @@
...
+//! [ac_re]: atspi-connection::AccessibilityConnection::register_event
+//! [ac_de]: atspi-connection::AccessibilityConnection::deregister_event
+//! [ac]: atspi-connection::AccessibilityConnection
```
TTWNO: None of the above three links work; just stay on the same page.

README.md
```diff
@@ -26,12 +26,18 @@ We use the asynchronous zbus API, so to use atspi, you will need to run an async
...
 
-- `default`: `proxies`, `connection`.
-- `proxies`: enable re-export of the `atspi-proxies` crate; this allows you to directly communicate with DBus.
-- `connection`: enable re-export of the `atspi-connection` crate; this gives some nice abstractions over DBus when receiving only. `proxies` will still be needed to query information actively.
-- `tokio`: enable support for the `tokio` runtime; other runtimes can be used without an integration feature.
-    - One exception on `glomio` as it has its own types not related to how other runtimes work; `atspi` is **_not_** compatible with `glomio` (PRs welcome though)
-- `tracing`: enable support for the `tracing` logger.
+| Flag | Default | Description |
+|------|:-------:|-------------|
+| `proxies` | ✅ | Re-exports the `atspi-proxies` crate, letting you actively communicate with (query) D-Bus. |
```
TTWNO: Use "yes"/"no" instead; blank cells don't always work well with screen readers.

TTWNO (changes requested):

luukvanderduim:
Note on the "missing" image: 

The local relative link to the image won't work because rustdoc does not copy the image to the target directory - nor will it adjust the path on the fly.
```rust
![Socket/Plug stitching sequence](./socket-plug.svg)
```
This is work:
```rust
![Socket/Plug stitching sequence](https://raw.githubusercontent.com/odilia-app/atspi/main/atspi-proxies/doc/socket-plug.svg)
```
If all is well, the svg will be shown after we publish.

TTWNO (approved):

PR #354, opened 2026-06-13
refactor(conn): DRYer and simpler `P2P` impl

luukvanderduim:
These methods in the `P2P` impl on `AccessibleConnection` can be written simpler and a little cheaper:

`bus_name_as_root_accessible`
`object_as_accessible`

In the lookup, When the peer is found, instead of cloning the `Peer` only clone the `zbus::Connection`, it is all we need.

Simplify by having only one builder in the method.


TTWNO:
*reduce, not `conduce`, LOL

TTWNO (approved):

PR #356, opened 2026-06-13
Update XML: `Role::Switch`

TTWNO:
- Update `xml/Accessible.xml`: add `Role::Switch`
- Upadte `xml/Application.xml`: add `getApplicationBusAddress` (already in Rust code)

### DIFF
--- a/atspi-common/xml/Application.xml
+++ b/atspi-common/xml/Application.xml
@@ -55,5 +55,13 @@
       <arg direction="out" type="s"/>
     </method>
 
+	<!-- Get the application's AT-SPI bus id to communicate directly with the application.
+
+	This allows for communication without being proxied by the daemon.
+	-->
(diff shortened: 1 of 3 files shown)


atspi-common/xml/Application.xml
```diff
@@ -55,5 +55,13 @@
       <arg direction="out" type="s"/>
     </method>
 
+	<!-- Get the application's AT-SPI bus id to communicate directly with the application.
```
luukvanderduim, line 58: I know this is probably what it says in XML but as a uninitiated reader, I would not know what the process is expected to return and what to do with it.

luukvanderduim (approved):
LGTM (just format and use conventional commit)

atspi-common/xml/Application.xml
TTWNO: Worth reporting it upstream.

Issue #357, opened 2026-06-16
Role serialisation in `Collection::get_matches`


luukvanderduim:
XML definitions state that the `MatchRule` "rule" has D-Bus type `(aiia{ss}iaiiasib)`

```XML
    <method name="GetMatches">
      <arg direction="in" name="rule" type="(aiia{ss}iaiiasib)"/>
      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QSpiMatchRule"/>
      <arg direction="in" name="sortby" type="u"/>
      <arg direction="in" name="count" type="i"/>
      <arg direction="in" name="traverse" type="b"/>
      <arg direction="out" type="a(so)"/>
      <annotation name="org.qtproject.QtDBus.QtTypeName.Out0" value="QSpiReferenceSet"/>
    </method>
```
in `atspi` `ObjectMatchRule` is defined as: 
```Rust
#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
pub struct ObjectMatchRule {
	pub states: StateSet,              // ai
	pub states_mt: MatchType,          // i
	pub attr: HashMap<String, String>, // a{ss}
	pub attr_mt: MatchType,            // i
	pub roles: Vec<Role>,              // ai
	pub roles_mt: MatchType,           // i
	pub ifaces: InterfaceSet,          // as
	pub ifaces_mt: MatchType,          // i
	pub invert: bool,                  // b
	// Private phantom, gets compiled away.
	// Here to ensure the builder is the only route to obtain a `MatchRule`
	#[serde(skip)]
	_marker: std::marker::PhantomData<()>,
}
```

Note that:
1. `Serialize` is derived
2. `roles` is defined as an array of `Role` enums.
3. `states` emloys the `StateSet` `Serialize` impl below the surface.

This means that even though we implemented our type according to XML, it will not serialize as 
a11y bus services expect: 

in at-spi2-core `libatspi` `MatchRule is defined as:
```C
typedef struct _AtspiMatchRule AtspiMatchRule;
struct _AtspiMatchRule
{
  GObject parent;
  AtspiStateSet *states;
  AtspiCollectionMatchType statematchtype;
  GHashTable *attributes;
  AtspiCollectionMatchType attributematchtype;
  GArray *interfaces;
  AtspiCollectionMatchType interfacematchtype;
  gint roles[5]; 
  AtspiCollectionMatchType rolematchtype;
  gboolean invert;
};
```
Solution: 
[...]

PR #358, opened 2026-06-17
fix(common): Introduce `RoleSet`

luukvanderduim:
Introduces `RoleSet` with all required mechanics.
It appears we cannot use `bitflags` on `RoleSet` because bitflags does not support an array as backing data.

`Serialize`, `Deserialize`, bitops, `IntoIterator`, `From<Role>` and tests.

Adjusted `ObjectMatchRule` and `MatchType` accordingly and/or to document the limitations found with `MatchType` in upstream atk-adaptor.

Addresses: issue #357

### DIFF
--- a/atspi-common/src/object_match.rs
+++ b/atspi-common/src/object_match.rs
@@ -3,7 +3,7 @@ use std::{borrow::Borrow, collections::HashMap, marker::PhantomData};
 use serde::{Deserialize, Serialize};
 use zvariant::{Signature, Type};
 
-use crate::{Interface, InterfaceSet, Role, State, StateSet};
+use crate::{Interface, InterfaceSet, Role, RoleSet, State, StateSet};
 
 /// Defines how an object-tree is to be traversed.
 /// Used in `CollectionProxy`.
@@ -30,13 +30,18 @@ pub enum TreeTraversalType {
 /// let builder = MatchRule::builder();
 /// ```
 ///
+/// # Note on Naming
+/// This struct is named `ObjectMatchRule` instead of `MatchRule` to avoid
+/// naming conflicts and confusion with [`zbus::MatchRule`], which is used for
+/// lower-level D-Bus message filtering.
+///
 #[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ObjectMatchRule {
 	pub states: StateSet,
 	pub states_mt: MatchType,
 	pub attr: HashMap<String, String>,
 	pub attr_mt: MatchType,
-	pub roles: Vec<Role>,
+	pub roles: RoleSet,
 	pub roles_mt: MatchType,
 	pub ifaces: InterfaceSet,
 	pub ifaces_mt: MatchType,
@@ -65,7 +70,7 @@ impl Type for ObjectMatchRule {
 		&Signature::I32,
 		<HashMap<&str, &str>>::SIGNATURE,
 		&Signature::I32,
-		<Vec<i32>>::SIGNATURE,
+		<RoleSet as Type>::SIGNATURE,
 		&Signature::I32,
 		<Vec<&str>>::SIGNATURE,
 		&Signature::I32,
@@ -83,17 +88,33 @@ impl ObjectMatchRule {
 
 /// The 'builder' type for `MatchRule`.
 /// Use its methods to set match criteria.
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct ObjectMatchRuleBuilder {
 	states: StateSet,
 	states_mt: MatchType,
 	attr: HashMap<String, String>,
 	attr_mt: MatchType,
-	roles: Vec<Role>,
+	roles: RoleSet,
 	roles_mt: MatchType,
 	ifaces: InterfaceSet,
 	ifaces_mt: MatchType,
 	invert: bool,
+}
+
+impl Default for ObjectMatchRuleBuilder {
+	fn default() -> Self {
+		Self {
+			states: StateSet::default(),
+			states_mt: MatchType::All,
+			attr: HashMap::default(),
+			attr_mt: MatchType::All,
+			roles: RoleSet::default(),
+			roles_mt: MatchType::All,
+			ifaces: InterfaceSet::default(),
+			ifaces_mt: MatchType::All,
+			invert: false,
+		}
+	}
 }
 
 impl ObjectMatchRuleBuilder {
@@ -101,10 +122,9 @@ impl ObjectMatchRuleBuilder {
 	#[must_use]
 	pub fn states<I>(mut self, states: I, mt: MatchType) -> Self
 	where
-		I: IntoIterator,
-		I::Item: Borrow<State>,
+		I: IntoIterator<Item = State>,
 	{
-		self.states = states.into_iter().map(|state| *state.borrow()).collect();
+		self.states = states.into_iter().collect();
 		self.states_mt = mt;
 		self
 	}
@@ -120,7 +140,7 @@ impl ObjectMatchRuleBuilder {
 	/// Insert a slice of `Role`s
 	#[must_use]
 	pub fn roles(mut self, roles: &[Role], mt: MatchType) -> Self {
-		self.roles = roles.into();
+		self.roles = RoleSet::new(roles.iter().copied());
 		self.roles_mt = mt;
 		self
 	}
@@ -164,28 +184,44 @@ impl ObjectMatchRuleBuilder {
 	}
 }
 
-/// Enumeration used by [`ObjectMatchRule`] to specify how to interpret [`ObjectRef`] objects.
+// `Empty` or `Invalid`
+// https://github.com/GNOME/at-spi2-core/blob/main/atk-adaptor/adaptors/collection-adaptor.c#L190-L214
+
+/// Enumeration used by [`ObjectMatchRule`] to specify how to select for [`ObjectRef`] objects.
+///
+/// This specifies the matching criteria for properties such as states, roles, interfaces, and attributes.
 ///
 /// [`ObjectRef`]: crate::object_ref::ObjectRef
+///
+// <https://github.com/GNOME/at-spi2-core/blob/main/atspi/atspi-constants.h#L203-L228>
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Type, Default)]
 #[repr(i32)]
 pub enum MatchType {
+	/// No objects match under this criterion.
+	///
+	/// *Note:* This will invalidate the entire `ObjectMatchRule`.
+	///
+	/// This represents an invalid or uninitialized match criterion.
 	#[default]
-	/// Invalidates match criterion. Meanting: the search of this property will not be performed.
-	Invalid,
+	Invalid = 0,
 
-	/// All of the criteria must be met.
-	All,
+	/// All objects that meet all of the specified criteria match.
+	///
+	/// If the set is empty, all objects match.
+	All = 1,
 
-	/// Any of the criteria must criteria must be met.
-	Any,
+	/// All objects that meet at least one of the specified criteria match.
+	Any = 2,
 
-	/// None of the criteria must be met.
-	NA,
+	/// All objects that meet none of the specified criteria match.
+	None = 3,
 
-	/// Same as [`Self::All`] if the criterion item is non-empty - All of the criteria must be met.
-	/// For empty criteria this rule requires the returned value to also have empty set.
-	Empty,
+	/// All objects that meet all criteria match if the criteria set is non-empty;
+	/// for empty criteria, only objects that also have an empty set for this property match.
+	///
+	/// *Note:* This is currently unimplemented in the upstream server-side C library,
+	/// behaving identically to `Invalid`, matching no objects for the entire `ObjectMatchRule`.
+	Empty = 4,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Type)]
@@ -274,7 +310,7 @@ mod tests {
 
 		assert_eq!(rule.states, StateSet::default());
 		assert_eq!(rule.attr, HashMap::new());
-		assert_eq!(rule.roles, Vec::new());
+		assert_eq!(rule.roles, RoleSet::default());
 		assert_eq!(rule.ifaces, InterfaceSet::default());
 		assert!(!rule.invert);
 	}
@@ -297,7 +333,7 @@ mod tests {
 			rule.attr,
 			[("name".to_string(), "value".to_string())].iter().cloned().collect()
 		);
-		assert_eq!(rule.roles, vec![Role::Alert]);
+		assert_eq!(rule.roles, RoleSet::new([Role::Alert]));
 		assert_eq!(rule.ifaces, InterfaceSet::new(Interface::Action));
 		assert!(rule.invert);
 	}
--- a/atspi-common/src/object_match.rs
+++ b/atspi-common/src/object_match.rs
@@ -184,8 +184,8 @@ impl ObjectMatchRuleBuilder {
 	}
 }
 
-// `Empty` or `Invalid`
-// https://github.com/GNOME/at-spi2-core/blob/main/atk-adaptor/adaptors/collection-adaptor.c#L190-L214
+// `Empty` or `Invalid` have no filter functions implemented and are ignored, invalidating the match rule.
+// <https://github.com/GNOME/at-spi2-core/blob/main/atk-adaptor/adaptors/collection-adaptor.c#L190-L214>
 
 /// Enumeration used by [`ObjectMatchRule`] to specify how to select for [`ObjectRef`] objects.
 ///
@@ -199,28 +199,32 @@ impl ObjectMatchRuleBuilder {
 pub enum MatchType {
 	/// No objects match under this criterion.
 	///
-	/// *Note:* This will invalidate the entire `ObjectMatchRule`.
+	/// *Note:* This will invalidate the entire `ObjectMatchRule`.\
+	/// There is no filter function implemented in the reference
+	/// implementation [`atk-adaptor`](https://github.com/GNOME/at-spi2-core/blob/main/atk-adaptor/adaptors/collection-adaptor.c#L190-L214).
+	/// This variant is ignored.
 	///
 	/// This represents an invalid or uninitialized match criterion.
-	#[default]
 	Invalid = 0,
 
-	/// All objects that meet all of the specified criteria match.
+	/// Objects meeting all of the specified criteria match.
 	///
 	/// If the set is empty, all objects match.
+	#[default]
 	All = 1,
 
-	/// All objects that meet at least one of the specified criteria match.
+	/// Objects meeting at least one of the specified criteria match.
 	Any = 2,
 
-	/// All objects that meet none of the specified criteria match.
+	/// Objects meeting none of the specified criteria match.
 	None = 3,
 
-	/// All objects that meet all criteria match if the criteria set is non-empty;
-	/// for empty criteria, only objects that also have an empty set for this property match.
+	/// No objects match under this criterion.
 	///
-	/// *Note:* This is currently unimplemented in the upstream server-side C library,
-	/// behaving identically to `Invalid`, matching no objects for the entire `ObjectMatchRule`.
+	/// # Unimplemented
+	/// This is unimplemented in the upstream server-side C library,
+	/// Use of this variant is ignored, just as `Invalid`, using `Empty` as `MatchType` in
+	/// an [`ObjectMatchRule`] will invalidate the match rule.
 	Empty = 4,
 }
 
--- a/atspi-common/src/object_match.rs
+++ b/atspi-common/src/object_match.rs
@@ -173,8 +173,6 @@ impl ObjectMatchRuleBuilder {
 /// This specifies the matching criteria for properties such as states, roles, interfaces, and attributes.
 ///
 /// [`ObjectRef`]: crate::object_ref::ObjectRef
-///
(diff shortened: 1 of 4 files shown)


atspi-common/src/object_match.rs
```diff
@@ -164,28 +184,44 @@ impl ObjectMatchRuleBuilder {
...
+// <https://github.com/GNOME/at-spi2-core/blob/main/atspi/atspi-constants.h#L203-L228>
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Type, Default)]
 #[repr(i32)]
 pub enum MatchType {
+	/// No objects match under this criterion.
+	///
+	/// *Note:* This will invalidate the entire `ObjectMatchRule`.
+	///
+	/// This represents an invalid or uninitialized match criterion.
 	#[default]
```
TTWNO: If the default invalidates the entire query, then it shouldn't be the default.

atspi-common/src/object_match.rs
```diff
@@ -164,28 +184,44 @@ impl ObjectMatchRuleBuilder {
...
-	/// Invalidates match criterion. Meanting: the search of this property will not be performed.
-	Invalid,
+	Invalid = 0,
 
-	/// All of the criteria must be met.
-	All,
+	/// All objects that meet all of the specified criteria match.
```
TTWNO: Remove the initial "All ..." on these comments; it is confusing because one of the variants is also called All.

The comment works without it: "Objects that meet _all_ the specified criteria."; "Objects that meet at least one of the specified criteria", etc.

atspi-common/src/object_match.rs
```diff
@@ -164,28 +184,44 @@ impl ObjectMatchRuleBuilder {
...
 
-	/// Same as [`Self::All`] if the criterion item is non-empty - All of the criteria must be met.
-	/// For empty criteria this rule requires the returned value to also have empty set.
-	Empty,
+	/// All objects that meet all criteria match if the criteria set is non-empty;
+	/// for empty criteria, only objects that also have an empty set for this property match.
+	///
+	/// *Note:* This is currently unimplemented in the upstream server-side C library,
+	/// behaving identically to `Invalid`, matching no objects for the entire `ObjectMatchRule`.
+	Empty = 4,
```
TTWNO, line 200: Is there a clippy lint to show up when people use this? Like "you are using a variant that is currently unsupported upstream".

Also, use a doc comment with a link that actually shows where the issue is. Otherwise it's just conjecture.

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
...
+
+	/// Checks if a specific [`Role`] is in the set.
+	#[must_use]
+	pub fn contains(&self, role: Role) -> bool {
+		let role_val = role as usize;
+		let index = role_val / 32; // index into the array
+		let bit = role_val % 32;
+		// residual: the bit position within the array element note that residual can never exceed 32
+		if let Some(&bits) = self.0.get(index) {
+			(bits & (1 << bit)) != 0
```
TTWNO: `(bits >> bit) & 0b1 == 0b1` more readable IMO

Compiler is probably smart enough to figure this one out either way.

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
...
+	}
+
+	/// Checks if a specific [`Role`] is in the set.
+	#[must_use]
+	pub fn contains(&self, role: Role) -> bool {
+		let role_val = role as usize;
+		let index = role_val / 32; // index into the array
+		let bit = role_val % 32;
+		// residual: the bit position within the array element note that residual can never exceed 32
+		if let Some(&bits) = self.0.get(index) {
```
TTWNO, line 662: The conditional is fine, but place an assertion here for debug code to catch: `assert!(index <= <MAX> / 32, "Invalid role index");`

This will catch if the role bounds begin to exceed the type size.

Maybe there is even a solid `const` API for defining some of these variables, like the max index for Roles, max /32 index? Might be a fun rabbit hole for you if you want a fun afternoon, lol!

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
...
+			*bits |= 1 << bit;
+		}
+	}
+
+	/// Removes a [`Role`] from the set.
+	pub fn remove(&mut self, role: Role) {
+		let role_val = role as usize;
+		let index = role_val / 32;
+		let bit = role_val % 32;
+		if let Some(bits) = self.0.get_mut(index) {
```
TTWNO, line 680: Likewise.

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
...
+			*bits &= !(1 << bit);
+		}
+	}
+
+	/// Toggles a [`Role`] in the set.
+	pub fn toggle(&mut self, role: Role) {
+		let role_val = role as usize;
+		let index = role_val / 32;
+		let bit = role_val % 32;
+		if let Some(bits) = self.0.get_mut(index) {
```
TTWNO, line 688: Likewise.

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
...
+pub struct RoleSetIter {
+	set: RoleSet,
+	index: usize,
+}
+
+impl Iterator for RoleSetIter {
+	type Item = Role;
+
+	fn next(&mut self) -> Option<Self::Item> {
+		while self.index < 160 {
```
TTWNO: No hardcoding max values. This guarantees future headaches. Find another way of determining the largest allowed number.

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
...
+impl std::ops::BitAndAssign for RoleSet {
+	fn bitand_assign(&mut self, other: Self) {
+		*self = *self & other;
+	}
+}
+
+#[derive(Clone, Debug)]
+pub struct RoleSetIter {
+	set: RoleSet,
+	index: usize,
```
TTWNO: The only reason to use usize here is that it's the default for interator types. If we know it fits in a smaller type (and we need to use that smaller type internally), then just use it: `u32`.

Even further: if we really are just iterating through Roles, why not just store the current role we're checking on here? Like `Role::Paragraph` instead of `52`?

atspi-common/src/role.rs
```diff
@@ -648,4 +896,96 @@ pub mod tests {
...
+		use zvariant::serialized::Data;
+
+		let mut set = RoleSet::empty();
+		set.insert(Role::Alert); // Role::Alert is index 2, bit 2 inside the first i32 element.
+		set.insert(Role::Button); // Role::Button is index 43, bit 11 inside the second i32 element (43 - 32 = 11).
+
+		let ctxt = Context::new_dbus(LE, 0);
+		let encoded = to_bytes(ctxt, &set).unwrap();
+
+		let expected_bytes = &[
```
TTWNO, line 1062: I assume an LLM wrote this; but it's such a cool test having all the bytes of the encoding scheme laid out so nicely like this!

atspi-common/src/state.rs
```diff
@@ -475,15 +475,10 @@ impl<'de> Deserialize<'de> for StateSet {
...
-				match <Vec<u32> as Deserialize>::deserialize(deserializer) {
-					Ok(states) if states.len() == 2 => {
-						let mut bits = u64::from(states[0]);
-						bits |= (u64::from(states[1])) << 32;
-						StateSet::from_bits(bits).map_err(|_| de::Error::custom("invalid state"))
-					}
-					Ok(states) => Err(de::Error::invalid_length(states.len(), &"array of size 2")),
-					Err(e) => Err(e),
-				}
+				let arr = <[u32; 2] as Deserialize>::deserialize(deserializer)?;
```
TTWNO, line 478: Good catch!

TTWNO (changes requested):

atspi-common/src/object_match.rs
luukvanderduim: Reading these comments again makes me realize we doc-comment the variants Empty and Invalid as-if they had real-world meaning instead of an ancient intention. I am going to find out about what Qt and Gtk4 do in these cases. If those follow `atk-adaptor`'s footsteps I am inclined to look for something akin to #[deprecated] for these variants.

atspi-common/src/state.rs
DataTriny, line 1: Changes to this file have nothing to do with this pull request, please move them somewhere else.

atspi-common/src/role.rs
```diff
@@ -648,4 +896,96 @@ pub mod tests {
 			);
 		}
 	}
+
+	#[test]
+	fn test_role_set_operations() {
```
DataTriny: Please split this test (as well as the followings) and give them descriptive names so it is easier to understand what is going on if one fail.

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
...
+/// The bitflag representation of all roles an object may have.
+pub struct RoleSet([i32; 5]);
+
+impl RoleSet {
+	/// Create a new `RoleSet` from a collection of roles.
+	pub fn new<I>(roles: I) -> Self
+	where
+		I: IntoIterator<Item = Role>,
+	{
+		let mut set = Self::empty();
```
DataTriny:
```suggestion
		Self::from_iter(roles)
```

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
 	}
 }
 
+//#[allow(clippy::module_name_repetitions)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Default)]
+/// The bitflag representation of all roles an object may have.
```
DataTriny, line 625: It's not idiomatic to place documentation after the derive macros.

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
 	}
 }
 
+//#[allow(clippy::module_name_repetitions)]
```
DataTriny: What's that?

atspi-common/src/role.rs
```diff
@@ -617,6 +622,249 @@ impl std::fmt::Display for Role {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Default)]
```
DataTriny: enumflags2 lets us display both the bits and the set variants for a `StateSet`'s debug representation, this might be desirable to have here too.

DataTriny (changes requested):

luukvanderduim:
@DataTriny It appears `Collection` has fallen out of favor? Does `AccessKit` implement it?

DataTriny:
> It appears Collection has fallen out of favor?

What makes you say that?

> Does AccessKit implement it?

Not yet, but we'll need it as Orca mandates it for its structural navigation feature to work, see https://github.com/accesskit/accesskit/issues/603.

atspi-common/src/role.rs
```diff
@@ -617,6 +622,255 @@ impl std::fmt::Display for Role {
...
+	/// Checks if all roles are unset.
+	#[must_use]
+	pub fn is_empty(&self) -> bool {
+		self.0.iter().all(|&bits| bits == 0)
+	}
+
+	/// Checks if a specific [`Role`] is in the set.
+	#[must_use]
+	pub fn contains(&self, role: Role) -> bool {
+		let role_val = role as usize;
```
DataTriny: These same lines are repeated multiple times throughout the file, they should likely be extracted into `fn byte_and_bit_for(role: Role) -> (usize, u32)` or something.

atspi-common/src/object_match.rs
```diff
@@ -164,28 +184,48 @@ impl ObjectMatchRuleBuilder {
...
-	NA,
+	/// Objects meeting none of the specified criteria match.
+	None = 3,
 
-	/// Same as [`Self::All`] if the criterion item is non-empty - All of the criteria must be met.
-	/// For empty criteria this rule requires the returned value to also have empty set.
-	Empty,
+	/// No objects match under this criterion.
+	///
+	/// # Unimplemented
```
DataTriny: Personally I wouldn't call out one implementation in the documentation.

atspi-common/src/object_match.rs
```diff
@@ -164,28 +184,48 @@ impl ObjectMatchRuleBuilder {
...
 #[repr(i32)]
 pub enum MatchType {
-	#[default]
-	/// Invalidates match criterion. Meanting: the search of this property will not be performed.
-	Invalid,
+	/// No objects match under this criterion.
+	///
+	/// *Note:* This will invalidate the entire `ObjectMatchRule`.\
+	/// There is no filter function implemented in the reference
+	/// implementation [`atk-adaptor`](https://github.com/GNOME/at-spi2-core/blob/main/atk-adaptor/adaptors/collection-adaptor.c#L190-L214).
```
DataTriny: The at-spi2-core project does not officially live on GitHub, fix this link if you think it is relevant. Keep in mind that the project has an incompatible license though, so you are not supposed to consult it while working here.

atspi-common/src/object_match.rs
```diff
@@ -164,28 +184,48 @@ impl ObjectMatchRuleBuilder {
...
+// `Empty` or `Invalid` have no filter functions implemented and are ignored, invalidating the match rule.
+// <https://github.com/GNOME/at-spi2-core/blob/main/atk-adaptor/adaptors/collection-adaptor.c#L190-L214>
+
+/// Enumeration used by [`ObjectMatchRule`] to specify how to select for [`ObjectRef`] objects.
+///
+/// This specifies the matching criteria for properties such as states, roles, interfaces, and attributes.
 ///
 /// [`ObjectRef`]: crate::object_ref::ObjectRef
+///
+// <https://github.com/GNOME/at-spi2-core/blob/main/atspi/atspi-constants.h#L203-L228>
```
DataTriny: Same comment as below, what information does this bring that couldn't be shown here?

atspi-common/src/object_match.rs
```diff
@@ -164,28 +184,48 @@ impl ObjectMatchRuleBuilder {
 	}
 }
 
-/// Enumeration used by [`ObjectMatchRule`] to specify how to interpret [`ObjectRef`] objects.
+// `Empty` or `Invalid` have no filter functions implemented and are ignored, invalidating the match rule.
```
DataTriny: If a combination of values truly make no sense, why not just making them impossible to represent in the type system and let zbus error out early? This pull request is about fixing role set though, so these changes are slightly out of scope in the first place and only make review longer.

DataTriny (changes requested):

luukvanderduim:
> > It appears Collection has fallen out of favor?
> 
> What makes you say that?

Not exhaustive research but I got that sense by looking around with [collection-check](https://github.com/luukvanderduim/collection-check). (A small helper I whipped up to help confirm whether our `Collection::get_matches` works.)

'Old' toolkits do provide Collection (on root) but Qt 6.9 or Gtk4 don't.

So actually I am very happy to learn Orca mandates it.

> 
> > Does AccessKit implement it?
> 
> Not yet, but we'll need it as Orca mandates it for its structural navigation feature to work, see [AccessKit/accesskit#603](https://github.com/AccessKit/accesskit/issues/603).

TTWNO (approved):

atspi-common/src/object_match.rs
```diff
@@ -164,28 +168,45 @@ impl ObjectMatchRuleBuilder {
...
 }
 
-/// Enumeration used by [`ObjectMatchRule`] to specify how to interpret [`ObjectRef`] objects.
+/// Enumeration used by [`ObjectMatchRule`] to specify how to select for [`ObjectRef`] objects.
+///
+/// This specifies the matching criteria for properties such as states, roles, interfaces, and attributes.
 ///
 /// [`ObjectRef`]: crate::object_ref::ObjectRef
+///
+// <https://github.com/GNOME/at-spi2-core/blob/main/atspi/atspi-constants.h#L203-L228>
```
DataTriny:
```suggestion
// <https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/main/atspi/atspi-constants.h?ref_type=heads#L203-228>
```

atspi-common/src/object_match.rs
```diff
@@ -164,28 +168,45 @@ impl ObjectMatchRuleBuilder {
...
 #[repr(i32)]
 pub enum MatchType {
-	#[default]
-	/// Invalidates match criterion. Meanting: the search of this property will not be performed.
-	Invalid,
+	/// No objects match under this criterion.
+	///
+	/// *Note:* This will invalidate the entire `ObjectMatchRule`.\
+	/// There is no filter function implemented in the reference
+	/// implementation [`atk-adaptor`](https://github.com/GNOME/at-spi2-core/blob/main/atk-adaptor/adaptors/collection-adaptor.c#L190-L214).
```
DataTriny:
```suggestion
	/// implementation [`atk-adaptor`](https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/main/atk-adaptor/adaptors/collection-adaptor.c?ref_type=heads#L190-214).
```

atspi-common/src/role.rs
```diff
@@ -617,9 +622,248 @@ impl std::fmt::Display for Role {
...
+#[derive(Clone, Debug)]
+pub struct RoleSetIter {
+	set: RoleSet,
+	index: usize,
+}
+
+impl Iterator for RoleSetIter {
+	type Item = Role;
+
+	fn next(&mut self) -> Option<Self::Item> {
```
DataTriny, line 810: Can be simplified:

```
diff --git a/atspi-common/src/role.rs b/atspi-common/src/role.rs
index f9aa61b..760e511 100644
--- a/atspi-common/src/role.rs
+++ b/atspi-common/src/role.rs
@@ -794,21 +794,17 @@ impl std::ops::BitAndAssign for RoleSet {
 #[derive(Clone, Debug)]
 pub struct RoleSetIter {
        set: RoleSet,
-       index: usize,
+       index: u32,
 }

 impl Iterator for RoleSetIter {
        type Item = Role;

        fn next(&mut self) -> Option<Self::Item> {
-               let bitfield_length = std::mem::size_of_val(&self.set.0) * (u8::BITS as usize);
-               while self.index < bitfield_length {
-                       let current = u32::try_from(self.index).ok()?;
+               while let Ok(role) = Role::try_from(self.index) {
                        self.index += 1;
-                       if let Ok(role) = Role::try_from(current) {
-                               if self.set.contains(role) {
-                                       return Some(role);
-                               }
+                       if self.set.contains(role) {
+                               return Some(role);
                        }
                }
                None
```

atspi-common/src/role.rs
```diff
@@ -617,9 +622,248 @@ impl std::fmt::Display for Role {
...
+}
+
+impl std::ops::BitAndAssign for RoleSet {
+	fn bitand_assign(&mut self, other: Self) {
+		*self = *self & other;
+	}
+}
+
+#[derive(Clone, Debug)]
+pub struct RoleSetIter {
```
DataTriny: You must implement `FusedIterator`, and preferably `ExactSizeIterator` too. I'd also rename this type to `RoleSetIterator` for better readability.

DataTriny (changes requested):
@luukvanderduim You've asked me for a new review without actually addressing my comments or justifying why not to. Please go through all of them, I am not going to approve this before then.

atspi-common/src/object_match.rs
```diff
@@ -164,28 +168,38 @@ impl ObjectMatchRuleBuilder {
...
 }
 
-/// Enumeration used by [`ObjectMatchRule`] to specify how to interpret [`ObjectRef`] objects.
+/// Enumeration used by [`ObjectMatchRule`] to specify how to select for [`ObjectRef`] objects.
+///
+/// This specifies the matching criteria for properties such as states, roles, interfaces, and attributes.
 ///
 /// [`ObjectRef`]: crate::object_ref::ObjectRef
+///
+// <https://github.com/GNOME/at-spi2-core/blob/main/atspi/atspi-constants.h#L203-L228>
```
DataTriny: This link wasn't fixed despite my previous suggestion.

atspi-common/src/role.rs
```diff
@@ -617,9 +622,267 @@ impl std::fmt::Display for Role {
...
+
+	/// Returns an iterator yielding each set [`Role`].
+	#[must_use]
+	pub fn iter(&self) -> RoleSetIterator {
+		RoleSetIterator { set: *self, index: 0, remaining: self.len() }
+	}
+
+	/// Returns the number of roles in this set.
+	#[must_use]
+	pub fn len(&self) -> u32 {
```
DataTriny: `len` methods always ought to return a usize. Please cast the sum of u32s.

atspi-common/src/role.rs
```diff
@@ -617,9 +622,267 @@ impl std::fmt::Display for Role {
...
+	set: RoleSet,
+	index: u32,
+	remaining: u32,
+}
+
+impl Iterator for RoleSetIterator {
+	type Item = Role;
+
+	fn next(&mut self) -> Option<Self::Item> {
+		if self.remaining == 0 {
```
DataTriny, line 811: I don't think this check is needed. Once past the last `Role` variant, `Role::from(u32)` will always return `None`.

DataTriny (dismissed):

luukvanderduim:
Thank you Tait and Arnold. You have raised valuable concerns and made helpful suggestions that improved this PR markedly. Almost all have been adopted or addressed. Both `RoleSet` and its implementations are in great shape now. To keep the momentum, I will dismiss open reviews, merge this now, and open a small follow-up PR for polishing.

Issue #360, opened 2026-06-18
Move uses of type `#[allow(clippy::lint)]` to `#[expect(clippy::lint, reason = "variable is unused when `feature` is not selected"]`


luukvanderduim:
Consider moving our Clippy-escapes (`#[allow(clippy::lint-name)]` to:

```Rust
#[expect(clippy::lint_name, reason = "False positive in Clippy v1.x")]
```

The built-in attribute `expect` is more elegant in that it will notify users when the decorated code no longer trips the lint. (for instance. because the lint was rewritten or the code changed which made the "unused variable" used).

[reference for expect attribute](https://doc.rust-lang.org/reference/attributes/diagnostics.html#r-attributes.diagnostics.expect)

PR #361, opened 2026-06-20
chore(common): Convert `allow` into `expect` attributes.

luukvanderduim:
# Convert `allow` to `expect` attribute.

The [`#[expect]` attribute](https://doc.rust-lang.org/reference/attributes/diagnostics.html#r-attributes.diagnostics.lint.expect) is smarter than `#[allow]`.

* `#[allow( lint )]` unconditionally suppresses the lint warning on the item.
* `#[expect( lint )]` suppresses the lint warning **if the lint warning is (still) relevant**
   otherwise it will say the lint is unfulfilled, in which case it can be removed.

This protects against stale `#[allow]` attributes.

Most `allow` attributes were found on the `module_repetition` lint, a few were found to be no longer relevant!
One `too_many_arguments` on a `Collection` method.
One `function_too_long` in p2p.

One `derivable_impl` on a manual `Default` implementation deserves a revisit, because it requires the type to change in order to remove the impl.

solves: #360


atspi-common/src/events/object.rs
```diff
@@ -59,8 +59,11 @@ impl Hash for PropertyChangeEvent {
 impl Eq for PropertyChangeEvent {}
 
 // TODO: Looks like a false positive Clippy lint
-// Derive me.
-#[allow(clippy::derivable_impls)]
+// Derive me after removing string from event.
+#[expect(
+	clippy::derivable_impls,
+	reason = "Cannot derive `Default` because `String::default` != `Property::default`"
```
TTWNO, line 65: Even if this is true, we are still defining it in these terms... fix the String default, instead of excusing the clippy lint which is telling us we did it wrong.

atspi-common/src/role.rs
```diff
@@ -336,142 +336,143 @@ pub enum Role {
 impl TryFrom<u32> for Role {
 	type Error = AtspiError;
 
-	#[allow(clippy::too_many_lines)]
+	#[expect(
```
TTWNO, line 339: See Arnold's PR. #363 potentially worth looking at for a way to remove this lint.

atspi-connection/src/p2p.rs
```diff
@@ -468,7 +468,7 @@ impl Peers {
 	///
 	/// # Note
 	/// This function is called internally by `AccessibilityConnection::new()`.
-	#[allow(clippy::too_many_lines)]
+	#[expect(clippy::too_many_lines, reason = "Nethod exceeds the maximum allowed line count.")]
```
TTWNO, line 471: This `fn` is pretty big; I'd rather listen to the lint here.

atspi-connection/src/p2p.rs
```diff
@@ -586,7 +586,7 @@ impl Peers {
 
 								// Well-known name received a new owner on the bus.
 								(Some(old_owner_unique_name), Some(new_owner_unique_name)) => {
-									if let Ok(()) = peers.update_well_known_owner(&well_known_name, old_owner_unique_name, new_owner_unique_name, &conn).await.inspect_err(|#[allow(unused_variables)] err| {
+									if let Ok(()) = peers.update_well_known_owner(&well_known_name, old_owner_unique_name, new_owner_unique_name, &conn).await.inspect_err(| #[cfg_attr(not(feature = "tracing"), expect(unused_variables))] err| {
```
TTWNO, line 589: Lines like this are getting pretty hard to read. Likewise with a few of the changes above. Maybe split this out into another function or something. There must be a way to avoid the `.x.y.z.awai.inspect_err(|#cfg(...) err| + closule)` from becoming pervasive here.

Cargo.toml
```diff
@@ -12,7 +12,7 @@ authors = [
   "Tait Hoyem <tait@tait.tech>",
 ]
 description = "AT-SPI2 protocol implementation in Rust"
-rust-version = "1.77.2"
+rust-version = "1.81.0"
```
TTWNO, line 15: Why are we bumping the version? Is this required purely for using `expect`?

It's fine if it is, 1.81 came out quite a while ago.

TTWNO (changes requested):
If you are going through all this effort, add `clippy::allow_attribute_without_reason` to the deny list.

Cargo.toml
luukvanderduim: From the commit message:

The `#[expect]` attribute - with reason - was stabilized in 1.81.0.

If everyone is on-board, we can also (finally) bump dev-dependency
`criterion` which, as it appears, waited for us to leap to >=1.80.0.

Let's see what CI thinks now.

PR #362, opened 2026-06-23
refactor(common): polish for role

luukvanderduim:
* Moved `element_idx_and_bit_for` into the `RoleSet` impl as associated function.

* Added `RoleSet::ARRAY_LEN`, the backing data eleement count, determined at compile time.

* Converted uses of `Self::empty().0.len()` to `Self::ARRAY_LEN`

This now makes `Serialize` drop a 'magic' number and be agnostic over backing data length.

### DIFF
--- a/atspi-common/src/role.rs
+++ b/atspi-common/src/role.rs
@@ -626,15 +626,10 @@ impl std::fmt::Display for Role {
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Default)]
 pub struct RoleSet([i32; 5]);
 
-fn element_idx_and_bit_for(role: Role) -> (usize, usize) {
-	let role_val = role as usize;
-	let index = role_val / 32;
-	debug_assert!(index < RoleSet::empty().0.len());
-	let bit = role_val % 32;
-	(index, bit)
-}
-
 impl RoleSet {
+	// Determines the number of elements in the backing array at compile time.
+	const ARRAY_LEN: usize = std::mem::size_of::<Self>() / std::mem::size_of::<i32>();
+
 	/// Create a new `RoleSet` from a collection of roles.
 	pub fn new<I>(roles: I) -> Self
 	where
@@ -646,7 +641,7 @@ impl RoleSet {
 	/// Create an empty `RoleSet`.
 	#[must_use]
 	pub const fn empty() -> Self {
-		Self([0; 5])
+		Self([0; RoleSet::ARRAY_LEN])
 	}
 
 	/// Checks if all roles are unset.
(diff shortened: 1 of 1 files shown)


TTWNO (approved):

DataTriny:
Actually I'd do it the other way around.

```rust
const ROLE_SET_SIZE: usize = f32::ceil(Role::PushButtonMenu as u32 as f32 / i32::BITS as f32) as usize;
pub struct RoleSet([i32; ROLE_SET_SIZE]);
```

Then put a comment near the `Role` enum to remind people to update this constant if roles are added or removed, like when you will rebase and get the `Switch` variant. No more magic 5, no debug assertions needed. Don't forget to update `RoleSetVisitor::expecting` which would now use the constant.

TTWNO:
> Actually I'd do it the other way around.
> 
> ```rust
> const ROLE_SET_SIZE: usize = f32::ceil(Role::PushButtonMenu as u32 as f32 / i32::BITS as f32) as usize;
> pub struct RoleSet([i32; ROLE_SET_SIZE]);
> ```
> 
> Then put a comment near the `Role` enum to remind people to update this constant if roles are added or removed, like when you will rebase and get the `Switch` variant. No more magic 5, no debug assertions needed. Don't forget to update `RoleSetVisitor::expecting` which would now use the constant.

Surely you don't believe this is a better solution? If you are suggesting static assertions vs runtime, that is understandable—perhaps there is something we could do there. That casting construction is unforgivable in basic struct declaration to calculate a number—the magic "5" is better than that, frankly... *and* your auggestion still requires changing code in multiple places.

luukvanderduim:
I'll add a test that will tell us whether the backing array has run out of space. That does not require `const fn` or recent rustc versions.

If @DataTriny is arguing we are being a little silly abstracting over the size of the backing store, he has a point. 
Considering there were four roles added since 2020 and `RoleSet` has room for 29 more. At current rate it will need a sixth element in about 43 years.

Rust has [a few features](https://doc.rust-lang.org/stable/std/mem/fn.variant_count.html) in the making that will make it trivial to couple the array length to the enum's size. I can do it now but as I see it, it would only add sillyness.

luukvanderduim:
Just call me when this trips ;)

```rust
	#[test]
	fn last_variant_fits_in_role_set() {
		let last_role_variant =
			(0..).into_iter().find(|&i| Role::try_from(i).ok().is_none()).unwrap() - 1;
		let index_in_array = last_role_variant / 32;
		assert_eq!(5, index_in_array + 1, "RoleSet's array has insufficient number of elements");
	}
```

PR #363, opened 2026-06-23
refactor(common): Define the Role enum with a macro

DataTriny:
Makes it easier to keep variants and their human-readable names in sync, automatically provides the `TryFrom` impl, allows defining a constant which points to the last variant.

We cannot directly do pattern matching on expressions in the `TryFrom` impl, but I have verified that LLVM produces a jump table for this version, so it is kept branchless.

`Role::last_variant` would almost always get optimized to the single last value in release builds, but this is not guaranteed on debug builds. Using `Role::LAST_VARIANT` ensures that this is done at compile-time regardless. Intentionally left unused for #362.

### DIFF
--- a/atspi-common/src/role.rs
+++ b/atspi-common/src/role.rs
@@ -8,288 +8,341 @@ use zvariant::Type;
 
 use crate::AtspiError;
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Type, Hash)]
-/// An accessible object role.
-///
-/// To think of it in terms of HTML, any semantic element likely has a corollary in this enum.
-/// For example: `<button>`, `<input>`, `<form>` or `<h4>`.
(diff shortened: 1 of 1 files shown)


luukvanderduim:
Makes it a lot tidier, removes boiler plate and couples the strings to the variants. (And as a bonus we cannot forget to expand the backing store in ~43 years.) 
Nice, well done.

TTWNO (dismissed):

DataTriny:
While investigating a completely unrelated failure of the semver-checks CI job I found that both your latest pull requests @luukvanderduim and @TTWNO introduced minor and breaking changes in atspi-common.

Since the checks run for the atspi crate on every commits and nothing is mentioned in the release checklist, I assume you forgot to bump the version on your respective pull requests...

Shall I merge this one myself?

DataTriny:
@TTWNO This one would need your approval again. Didn't plan to use the constant initially but found it could be useful in the test suite.

TTWNO (approved):

PR #365, opened 2026-06-27
Use workspace semver check

TTWNO:
Previously, we were only testing the `atspi` crate itself for its public API.
Since much of the code lives in subcrates, recent bump requirements were completely missed—see #363, #356—and would continue to.

This instead checks semver for all crates in the workspace via the `--workspace` flag, and removes the comment about system TLS because it only used to be an option, but is no longer.

NOTE: I am expecting this to fail the semver check, as the already-merged PRs break `atspi-common`'s contract on crates.io.


TTWNO:
Huzza! That is the error Arnold found. Should never have slipped past CI—or our minds for that matter.

DataTriny (approved):
That's good. Versions should be bumped in a separate pull request. Might be worth merging it before this one so we can prove we won't break semver.

PR #366, opened 2026-06-27
Release 0.31.0

TTWNO:
Bump versions to match breaking changes.

- #365 found that `atpsi-common` had broken semver and needed a bump.
- This PR does the bump, including bumping subcrates to match atspi.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,7 +205,7 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "atspi"
-version = "0.30.0"
+version = "0.31.0"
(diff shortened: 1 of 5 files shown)


DataTriny:
So you want to have all the crates in sync, essentially released as a single unit? I know other projects do that but it needs to be documented as subcrates won't necessarily follow semver. Personally I don't have a preference. If that is what you want though, you should put the version only once in the workspace manifest and import it everywhere else.

TTWNO:
Thought about it more and decided to keep the current system of separate releases.

TTWNO:
Sorry, forgot to rebase again.

DataTriny (approved):
Changes look good. Less sure about the commit title though, it doesn't actually release anything but just bumps the crates versions.

Issue #368, opened 2026-06-27
Reconfigure CI Pipeline


TTWNO:
I noticed while writing #367 that the P2P tests run at the first stage of the pipeline. Which is too much.

Keep the heavy jobs at the end. Maybe I'll sort them by time and just group the <1m, 1-2m, >2m in their own stages.

We may also want to consider moving some of the tests to only run with a bot. The way the Rust compiler does this with bors is cool; you only do a full test run once everything is approved and ready to go.

luukvanderduim:
In current CI a few things stand out:

- We cache dependencies but do not cache build artifacts
  This means we need to recompile all dependencies from scratch each job.

- A number of jobs do not use caching at all.
  * `msrv-compliance`, 
  * `wasm-compatible-common-crate`, 
  * `atspi-common`,
  * `coverage`, 
  * `run-test-atspi-p2p` 

If we were to cache build artifacts manually with current caching, we would quickly exceed the 10GB limit of the runner. [`rust-cache`](https://github.com/swatinem/rust-cache) can help here because it applies `cargo-sweep` before saving, storing the relevant parts only.

The trade-off of storing `target` is that it tends to be large and has to be downloaded on each run that uses it. So it is advised to only use this on jobs that do compilations.

I've setup a [branch that keeps a cache of both deps and build artifacts](https://github.com/odilia-app/atspi/compare/cache-dependencies-and-build-artifacts-in-ci?expand=1) but avoids using artifacts in tasks thaat do not require those.

DataTriny:
One advantage of using a tool that manages a release pull request such as release-plz is that we only need to run cargo-semver-checks on those pull requests. These jobs are quite long and currently run for all pushes to all pull requests.

DataTriny:
I also don't understand why there is a job solely to check that the crates can compile with the declared MSRV. We could just use the MSRV on all jobs to begin with. This project doesn't rely on cutting edge Rust features and new Rust versions never break old code using the same edition.

TTWNO:
That is a good point, Arnold. Then we can cache with even greater efficiency.

luukvanderduim:
> I also don't understand why there is a job solely to check that the crates can compile with the declared MSRV. We could just use the MSRV on all jobs to begin with. This project doesn't rely on cutting edge Rust features and new Rust versions never break old code using the same edition.

That sounds good in principal, but..
Wouldn't that require pulling and installing that particular MSRV toolchain for each job?
[...]

PR #373, opened 2026-06-30
ci: Cache dependencies and build artifacts

luukvanderduim:
Before this PR:

- We cache dependencies for a number of jubs but we do not cache build artifacts.
- This means we need to recompile all dependencies from scratch each job.

- A number of jobs do not use dependency caching either:
  * `msrv-compliance`,
  * `wasm-compatible-common-crate`,
  * `atspi-common`,
  * `coverage`,
  * `run-test-atspi-p2p`

If we were to cache build artifacts manually with current generic caching, we would quickly exceed the 10GB limit of GitHub Actions. 
[`rust-cache`](https://github.com/swatinem/rust-cache) can help here because it applies `cargo-sweep` before saving build artifacts in cache, storing only the (still) relevant parts.

The trade-off of storing `target/` is that it tends to be large and has to be downloaded for each job that uses it. So it is advised to only use this cache on jobs where that trade-off is faborable, in ones that do compilations.

Not all jobs are compiling code, some only inspect dependencies. Here we can omit the `target/` cache.

This commit modifies the CI config:

with `target/` cache)
* `benchmarks`, `clippy`, `tests`, `rustdoc`, `semver-compliance`

Replaced manual `actions/cache` with `swatinem/rust-cache` + `shared-key: "atspi-shared-cache"`.

no `target/` cache:
* `cargo-deny`, `no-unused-dependencies` Replaced manual cache with `swatinem/rust-cache` + `cache-targets: "false"` to only cache downloaded registry indices.

'heavy' jobs that did not have caching:
* `msrv-compliance`, `wasm-compatible-common-crate`, `coverage`, `run-test-atspi-p2p` Added `swatinem/rust-cache` to previously uncached compiler-heavy jobs. `run-test-atspi-p2p`: cached both directories using the `workspaces` input.

Deliberately omitted:
* `find-msrv`, `rustfmt`, `typos`, `check-true-minimal-versions`

Addresses: #368

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,48 +23,35 @@ jobs:
     steps:
       - name: Git checkout
         uses: actions/checkout@v6
-      - name: Cache cargo home
-        uses: actions/cache@v5
-        env:
-          cache-name: cache-cargo-home
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@master
         with:
-          path: |
-            ~/.cargo/bin
-            ~/.cargo/registry/index
-            ~/.cargo/registry/cache
-            ~/.cargo/git/db
-          key: ${{ runner.os }}-x86_64-unknown-linux-gnu-build-${{ env.cache-name }}-${{ hashFiles('Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-x86_64-unknown-linux-gnu-build-${{ env.cache-name }}-
+          toolchain: stable
+      - name: Rust cache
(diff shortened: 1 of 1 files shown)


luukvanderduim:
Sorry, had to raise the branch to a PR to see whether the caching modifications mattered for the total actions run-time.

CI #1242:  4m 46s This branch
CI #1241:  6m 34s 
CI #1240:  5m:46s Introduced cargo-semver-checks
CI #1238:  4m 21s 

Either variability is very high or we have won what we lost with cargo-semver-checks.

PR #375, opened 2026-07-02
ci: Trim `aspi-test-p2p` job 

luukvanderduim:
Streamline `atspi-test-p2p`.
Observation: job `test-atspi-p2p` has a long run-time.

This PR applied:

- Trim dependencies downloaded.
- Do not install apt-recommended packages.
- Do not listen for tcp connections (these time out)
- Poll completion instead of blindly sleep.
- Sleep less.
- Cleanups.

Before: ~ 72 seconds
After:  ~ 39 seconds

Tested locally with 'act' ci runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -282,27 +282,26 @@ jobs:
           workspaces: |
             main
             test-atspi-p2p
-      - name: Install Dependencies
+      - name: Install System Dependencies
         run: |
           sudo apt-get -y update
-          sudo apt-get -y install \
+          sudo apt-get -y install --no-install-recommends -y \
             at-spi2-core \
-            libatspi2.0-dev \
             xvfb \
             dbus-x11 \
-            libgtk-3-dev \
-            gir1.2-gtk-3.0 \
-            mate-calc \
-            gnome-calculator \
-            metacity \
             gedit \
-            caja \
-            gsettings-desktop-schemas
+            metacity \
+            caja
       - name: Setup virtual display and AT-SPI environment
         run: |
-          # Start virtual display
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -282,27 +282,26 @@ jobs:
...
-          # Start virtual display
-          Xvfb :99 -screen 0 1024x768x24 &
-          sleep 2
+          # Start virtual display and wait until it's ready (max 10s)
+          Xvfb :99 -screen 0 1024x768x24 -nolisten tcp &
+          for i in $(seq 1 10); do
+            xdpyinfo -display :99 >/dev/null 2>&1 && break
+            sleep 0.1
+          done
+          # Ensure Xvfb is truly stable for subsequent GUI apps
```
TTWNO, line 303: What is this extra sleep for? If the `xdpyinfo` command resolves, why do we need to wait again? Not clear why.

If you really want this logic, put the sleep at the end instead of the beginning of the loop.

.github/workflows/ci.yml
```diff
@@ -312,37 +311,38 @@ jobs:
...
-          sleep 2
+          sleep 0.5
 
-          # Start AT-SPI bus and registry the standard way
+          # Start AT-SPI bus and registry the standard way, then wait for it (max 15s)
           /usr/libexec/at-spi-bus-launcher --launch-immediately &
-          sleep 5
-
-          # Explicitly enable accessibility
-          gsettings set org.gnome.desktop.interface toolkit-accessibility true
```
TTWNO, line 326: If the tests still pass with this disabled, then removing it is fine.

TTWNO (approved):

PR #376, opened 2026-07-02
ci: Integrate P2P test application

luukvanderduim:
Move `atspi-test-p2p` from Luuk's GitHub to the atspi workspace for maintainability.
Storing the test-application out of tree kept it out of maintenance.

- Modify the test application to depend on the local `atspi` tree.
- Adjust the test application to sleep less.
- No longer accept command line arguments and drop `argh` dependency.
- Rewrite the `README.md`
- Add the crate to the `atspi` workspace in a new `e2e` folder for end-to-end tests.

Adjust `ci.yml` accordingly.

Tested locally with 'act' local ci runner.
Reduced run-time of `test-atspi-p2p` to 35-39 seconds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members  = ["atspi", "atspi-common", "atspi-connection", "atspi-proxies"]
+members  = ["atspi", "atspi-common", "atspi-connection", "atspi-proxies",  "e2e/atspi-test-p2p"]
 resolver = "2"
 
 
--- a/e2e/atspi-test-p2p/README.md
+++ b/e2e/atspi-test-p2p/README.md
@@ -1,0 +1,54 @@
+# test-atspi-p2p
+
+The P2P feature of [`atspi`](<http://github.com/odilia-app/atspi.git>) spawns a task that listens for [`NameOwnerChanged`](<https://dbus.freedesktop.org/doc/dbus-specification.html#bus-messages-name-owner-changed>) events.
+
+If feature 'p2p' is selected, this `DBus` signal is used to update the `Peers` list on the `AccessibilityConnection`.
+This list of available P2P-capable, accessible applications is kept to provide users a P2P connection / proxy of the  
+application when they need it.
+
+## What it does
+
+1. Subscribes to tracing messages
+2. Shows the applications in the `Peers` list
+3. Launches a P2P capable application and verifies its appearance in the `Peers` list,
+   indicating the `NamoOwnerChanged` signal was received and handled as intended.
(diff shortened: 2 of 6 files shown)


e2e/atspi-test-p2p/README.md
```diff
@@ -0,0 +1,55 @@
+# test-atspi-p2p
+
+The P2P feature of [`atspi`](<http://github.com/odilia-app/atspi.git>) spawns a task that listens for [`NameOwnerChanged`](<https://dbus.freedesktop.org/doc/dbus-java/api/index.html?org/freedesktop/DBus.NameOwnerChanged.html>) events.
```
TTWNO: Use https://dbus.freedesktop.org/doc/dbus-specification.html#bus-messages-name-owner-changed

Why was the Java doc used?

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members  = ["atspi", "atspi-common", "atspi-connection", "atspi-proxies"]
+members  = ["atspi", "atspi-common", "atspi-connection", "atspi-proxies",  "e2e/atspi-test-p2p"]
```
TTWNO, line 2: I was wanting to check: this adds it only to the dev env and is never uploaded to crates.io.

But indeed only the content within the `atspi*/` directories actually gets published.

I understand why this is not part of the standard testing framework: all the setup necessary would more test the environment instead of the library itself. That said, should this be in `atspi/tests/` instead of its current top-level directory?

luukvanderduim: > I was wanting to check: this adds it only to the dev env and is never uploaded to crates.io.

You do not publish the workspace, you publish separate crates.
And if you were to try, it would fail because it has `publish = false` in `Cargo.toml`.

luukvanderduim: > That said, should this be in atspi/tests/ instead of its current top-level directory?

If we were to put the crate in `atspi/tests` that would be awkward at best.
If we were to extract the code and put it in `atspi/tests` it would become flaky because we cannot control what environment it is run in. 
If, for instance, a package integrator runs `cargo test`, they expect these to succeed or fail but not that success depends on a particular application being installed. Also it would require us to run _all_ our tests in the prepared environment because this particular test requires it.

As we discussed briefly on Matrix, there  is an opportunity here. If we have end-to-end tests, we can automate real world integration tests. Have real Gecko, Gnome, Qt or AccessKit back-ends emit signals and validate reception on our end.

e2e/atspi-test-p2p/README.md
luukvanderduim: Thanks for supplying the correct reference. I copied the wrong one from the original location.

PR #377, opened 2026-07-05
Make zbus-lockstep a dev dependency

DataTriny:
I try to be mindful of the number of dependencies and the time it take to compile accesskit_unix. zbus-lockstep is two crates, one procedural macro. That procedural macro even reaches for external files which can be considered an abuse of procedural macros.

This pull request allows downstream users such as AccessKit to disable this dependency, while keeping it enabled by default.

### DIFF
--- a/atspi-common/Cargo.toml
+++ b/atspi-common/Cargo.toml
@@ -23,8 +23,6 @@ enumflags2           = "0.7.9"
 serde                = "1.0.200"
 static_assertions    = "1.1.0"
 zbus                 = { workspace = true, optional = true }
-zbus-lockstep        = { version = "0.5.2" }
-zbus-lockstep-macros = { version = "0.5.2" }
 zbus_names           = "4.1.1"
 zvariant             = { version = "5.2", default-features = false }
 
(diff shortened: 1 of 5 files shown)


luukvanderduim:
After a bit of thought I think you are on the right track. 
You have found a downside to `zbus-lockstep-macros` friendliness, users are notified at compile-time about a mistake in an argument, this requires referencing the XML files.

But we can improve by doing:

```rust
#[cfg_attr(test, validate(signal: "SomeType"))]
```
This allows us to move the `zbus-lockstep` crates to `[dev-dependencies]` to avoid burden  AccessKit's users and make it a pure test-time concern.

This avoids the need for a new feature. 
New features double the power-set of the feature test in the `cargo-hack` CI job and why introduce one if there is a simpler alternative.

luukvanderduim:
Now documented in [zbus-lockstep](https://github.com/luukvanderduim/zbus-lockstep#test-gating-validate).

luukvanderduim (approved):
Nice. A win for everyone! Thanks!

PR #378, opened 2026-07-08
build (common): Require `zbus-lockstep` >= v0.6.0

luukvanderduim:
Addresses [ RUSTSEC-2026-0194 ](https://rustsec.org/advisories/RUSTSEC-2026-0194.html)

Also drop `zbus-lockstep-macros` from `[dev-dependencies]` using `macros` feature on `zbus-lockstep`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,39 +4,24 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.14.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "a55f82cfe485775d02112886f4169bde0c5894d75e79ead7eafe7e40a25e45f7"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
-name = "adler2"
-version = "2.0.1"
+name = "adler"
+version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
-
-[[package]]
-name = "aho-corasick"
-version = "1.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
-dependencies = [
- "memchr",
-]
+checksum = "6aa100a6f6f525226719f8de3f70076be4f4191801ebd92621450d1c51e9053d"
 
 [[package]]
 name = "anes"
-version = "0.1.6"
+version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


TTWNO:
We will need to bump MSRV with this PR as well: `1.77.2 -> 1.88`

@DataTriny `zbus` requires this. Nothing we can do about it.

luukvanderduim:
Hmm. `cargo msrv` on `zbus-lockstep` reports it requires "1.87.0" (Released May 15th 2025).

We don't really need to account for requirements imposed by `dev-dependencies` I think. Users should not need to care.

`cargo msrv find -- --lib` should help identify better which MSRV we should advertise.
That would allow us to use late(st) dev-deps while being considerate to users.

@TTWNO can you try that please and see where we end up?

TTWNO:
> Hmm. cargo msrv on zbus-lockstep reports it requires "1.87.0" (Released May 15th 2025).

Yes. Mistyped that. However, 1.87 is still the new MSRV for zbus; it's in the workspace Cargo.toml

DataTriny:
This pull request is missing the diff for Cargo.lock.

luukvanderduim:
I (try) to make it a habit to only push files I changed to avoid accidents.
But you are right - a fresh new minimal lockfile should be incuded.

DataTriny:
- I guess Cargo.lock is not properly managed in this project, otherwise there wouldn't be that much changes just to update a couple dependencies.
- In a pull request to bump a dependency version you remove unrelated code, couldn't you just update zbus_xml?
- Why have you included cosmetic changes to files that you haven't modified?

TTWNO:
> I guess Cargo.lock is not properly managed in this project, otherwise there wouldn't be that much changes just to update a couple dependencies.

Stop making sweeping generalizations about the project; you have done this numerous times. You continue to say things like "X is poorly managed/stupid/bad". This is not constructive to the project or me and Luuk who are volunteering to do this. In addition, there has been a decent number of times you've said something like this, and then I've had to correct you. Slow down and double check before you make these broad strokes.

`zbus` is the source of the vast majority of our dependencies (>100). Bumping it will necessarily bumps tons of transitive dependencies. Unless Luuk did `cargo update` and that just bumps everything (I doubt this is what happened though).

Please provide only constructive criticism going forward.

> Why have you included cosmetic changes to files that you haven't modified?

This is a good point, thank you. Luuk, please keep things to one change at a time. A formatting PR for Cargo files are fine, but especially deps that weren't touched, it doesn't make sense in the same PR. Makes it hard to sort out bisections later.

> In a pull request to bump a dependency version you remove unrelated code, couldn't you just update zbus_xml?

What code is being removed? I don't see it at all in the diff.

luukvanderduim:
I added the lockfile after reading [Change in Guidance on Committing Lockfiles (Rust blog)](https://blog.rust-lang.org/2023/08/29/committing-lockfiles/)

I think providing a minimal version with `cargo generate-lockfile -Zminimal-versions` is a proper way to make sure even the most conservative users have a tested library.
"not properly managed" is a qualification I cannot share.

Did I justify the equal signs, yes. Did rustfmt remove stale whitespace, yes it did.

The changes I made are relevant and warranted in my opinion.

PR #381, opened 2026-07-13
feat(conn)!: discover P2P peers lazily

serge-medvedev:
`AccessibilityConnection::new()` can hang before returning because P2P initialization probes and connects to every accessible application sequentially. One unresponsive application can therefore block the entire connection, even though P2P is optional and the accessibility bus remains usable.

This PR avoids that failure mode by discovering a peer on its first P2P-aware lookup. Construction now only installs a `NameOwnerChanged` listener so cached identities and connections can be invalidated when ownership changes.

Discovery state is kept by unique bus owner. Concurrent lookups for the same owner share one attempt, and stale attempts cannot publish after an ownership change. Unsupported applications are cached until invalidation, while transient address and connection failures use bounded per-owner backoff.

When no direct connection is available, `object_as_accessible` and `bus_name_as_root_accessible` continue over the existing accessibility-bus connection. Failures that prevent resolving or routing the target remain errors.

This changes two public methods:

- `get_peer` becomes async and returns `AtspiResult<Option<Peer>>`;
- `peers` returns a detached `Vec<Peer>` instead of `Arc<Mutex<Vec<Peer>>>`.

The latter prevents callers from locking or mutating discovery state. These changes are semver-relevant; maintainer guidance on release handling is welcome.

Tested with:

```text
cargo test -p atspi-connection --all-features
```

All 20 unit tests and 10 doctests pass.

Development and test design were assisted by `GPT-5.6 Sol`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,6 +252,7 @@ dependencies = [
  "atspi-common",
  "atspi-proxies",
  "enumflags2",
+ "event-listener",
  "futures-lite",
  "tokio-test",
  "tracing",
--- a/atspi-connection/Cargo.toml
+++ b/atspi-connection/Cargo.toml
@@ -15,13 +15,14 @@ version                = "0.15.0"
 
 [features]
 default  = ["p2p", "wrappers"]
-p2p      = ["dep:futures-lite", "zbus/p2p"]
+p2p      = ["dep:event-listener", "dep:futures-lite", "zbus/p2p"]
 tracing  = ["dep:tracing"]
 wrappers = ["atspi-common/wrappers", "dep:futures-lite"]
 
 [dependencies]
 atspi-common   = { path = "../atspi-common/", version = "0.15.0", default-features = false, features = ["zbus"] }
 atspi-proxies  = { path = "../atspi-proxies/", version = "0.15.0" }
+event-listener = { version = "5.4.1", optional = true }
 futures-lite   = { version = "2.6.0", default-features = false, optional = true }
 tracing        = { optional = true, workspace = true }
 zbus.workspace = true
--- a/atspi-connection/README.md
+++ b/atspi-connection/README.md
@@ -11,7 +11,7 @@ See the examples folder and documentation on how to use this library.
 ## Feature Flags
 
 * `default`: `wrappers`, `p2p`
-* `p2p`: dependencies `async-executor" and enables`zbus/p2p`
+* `p2p`: enables lazy peer discovery and `zbus/p2p`
 * `tracing`: enable support for the `tracing` crate
 * `wrappers`: enable support for `atspi-common` wrapper types that categorize events by interface, as well as the all-encompassing `Event` enum that can store any event type.
   * This also enables the `event_stream` function that allows you to receive a stream of `Event`s instead of specific events.
--- a/atspi-connection/p2p.md
+++ b/atspi-connection/p2p.md
@@ -8,8 +8,7 @@ The `Application` interface offers `GetApplicationBusAddress` which returns a bu
 
 ## `AccessibilityConnection`
 
-To avoid querying each object whether their bus name supports P2P, `AccessibilityConnection` keeps a list of all accessible applications on the bus that support P2P and continuously updates it.
-`AccessibilityConnection` is always in scope when performing operations on the accessibility bus, therefore it is the perfect place to keep the list of peers.
+`AccessibilityConnection` discovers an application's direct address only when a P2P-aware lookup first targets that application. Ready connections are reused, concurrent lookups for the same unique owner share one attempt, and normal P2P unavailability transparently falls back to the connection's long-lived accessibility bus.
 
 In practice, if you want to perform a method call on an `ObjectRef`, just get the `AccessibleProxy` for that object with `object_as_accessible`:
 
@@ -22,14 +21,13 @@ let name = obj_ap.name().await?;
 
 For `atspi` users who do not perform method calls or query properties, P2P is gated behind the "p2p" feature. Those users will need to opt out of default features.
 
-The feature "p2p" is enabled by default and if enabled, the `AccessibilityConnection` gains a list of applications that support P2P communication on initialization.
-Initialization of an `AccessibilityConnection` will also spawn a task to continuously listen for new applications entering or leaving the bus. It does so by listening for the `NameOwnerChanged` event emitted by the D-Bus daemon.
+The feature "p2p" is enabled by default. Initialization does not enumerate applications, request direct addresses, or open direct connections. It only starts a `NameOwnerChanged` listener used to invalidate cached identities; ownership signals never trigger discovery themselves.
 
-If users opt out, no list of `Peer`s will be kept and atspi will not listen for updates.
+If users opt out, no peer discovery state is kept and atspi does not start this listener.
 
 ## traits `P2P` and `Peer`
 
-As stated before, a list of `Peer`s is kept by the `AccessibilityConnection`.
+Ready peers are kept privately by `AccessibilityConnection`.
 
 The `Peer` struct can be considered a handle to individual peers that do support P2P and allows:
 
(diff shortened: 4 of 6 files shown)


TTWNO:
I appreciate this PR.

Ome thing I'm unsure of, is why the changes are so significant? Lazily adopting peers should not change _that much_ code in my opinion.

Can you please make a demonstration of the failure mode? It would be good for confirmatiion.

luukvanderduim:
Have you encountered hangs on creating a connection as user? 
Can you identify what application was unresponsive and perhaps why?

serge-medvedev:
> Have you encountered hangs on creating a connection as user? Can you identify what application was unresponsive and perhaps why?

@luukvanderduim , it was VS Code with Cline extension on Arch Linux. It's worth mentioning that I discovered the unwanted behavior while using `cua-driver` and ended up here.

TTWNO:
> it was VS Code with Cline extension on Arch Linux

VS code has occasionally been hard to deal with, so this makes sense. Thank you for clarifying.

> It's worth mentioning that I discovered the unwanted behavior while using cua-driver and ended up here.

Interesting. I had no idea they used our library. Thanks for letting us know.

serge-medvedev:
> Ome thing I'm unsure of, is why the changes are so significant? Lazily adopting peers should not change _that much_ code in my opinion.

Optimization suggestions are very welcome.

luukvanderduim:
This is interesting.

Do I understand correctly that 'Cline' using the 'cua-driver' has trouble initializing an `AccessibilityConnection` and that it is not necessarily VSCode itself that is not responding?

1. What I would like to know is which application is not responding to the `GetApplicationBusAddress` method call during initialization.
What is lazy initialization going to achieve if an application is not responding, right?
You only get to avoid, not solve the non-responding app with lazy initialization.

2. Is the application exclusively non-responsive on `GetApplicationBusAddress` or does it not respond to other calls either? 

The simplest way forward would be time-out after a few millis, preferably without introducing new dependencies.

If it were VSCode itself that is not responding, could you try 

```bash 
export ACCESSIBILITY_ENABLED=1
```
and restart VSCode?

Personally I think a few millis extra on initialization to handle a rude bus-server is not a problem.

luukvanderduim:
If we can, we should probably just use [Builder::method_timeout](https://docs.rs/zbus/latest/zbus/blocking/connection/struct.Builder.html#method.method_timeout) and perhaps allow users to configure it.

We would need to set it on all `Connection` builders, both the private (p2p) connections and the shared connection.

Then we close that failure mode entirely, hopefully _and_ it is a MUCH simpler fix. ;)

serge-medvedev:
> Do I understand correctly that 'Cline' using the 'cua-driver' has trouble initializing an `AccessibilityConnection` and that it is not necessarily VSCode itself that is not responding?

@luukvanderduim , not quite.

VS Code with Cline is running on a host with `cua-driver` installed and being used in the MCP server mode by an agent to drive the GUI.

serge-medvedev:
@TTWNO

> Can you please make a demonstration of the failure mode?

I ran a program whose only operation is `AccessibilityConnection::new()` under a 10-second external timeout, against both revisions in the same active desktop session:

- upstream `main`: timed out (`124`)
- this branch: returned in about 2.4 ms

@luukvanderduim

> What I would like to know is which application is not responding to the `GetApplicationBusAddress` method call during initialization.
>
> Is the application exclusively non-responsive on `GetApplicationBusAddress` or does it not respond to other calls either?

In this reproduction, the first registry entry was `:1.0`, identified as `xfce4-session`. It answered the accessibility-interface queries and `GetApplicationBusAddress`, returning an advertised Unix socket. Construction stalled only after that address had been returned, while creating the peer. A direct introspection request to the advertised socket also did not return within five seconds.

So my earlier identification of VS Code as the unresponsive application was too strong: VS Code was involved in the original user operation, but this minimal reproduction points to `xfce4-session`. `GetApplicationBusAddress` itself is not the call hanging here. I have not yet isolated the exact await inside peer creation, so I cannot say whether the stall is during P2P connection/authentication or a subsequent method call.

> If we can, we should probably just use `Builder::method_timeout` [...] Then we close that failure mode entirely, hopefully and it is a MUCH simpler fix.

That is an appealing, much smaller option. One caveat is that zbus applies `method_timeout` to method calls after a `Connection` has been built; the socket connection and authentication in `Builder::build()` happen before the timeout is installed on the connection. Therefore it would cover the failure if the stalled operation is a method call, but may not cover a stall during P2P connection/authentication. We need to isolate that exact step before claiming either approach closes the whole failure mode.

luukvanderduim:
> @TTWNO
> 
> > Can you please make a demonstration of the failure mode?
> 
> I ran a program whose only operation is `AccessibilityConnection::new()` under a 10-second external timeout, against both revisions in the same active desktop session:
> 
>     * upstream `main`: timed out (`124`)
> 
>     * this branch: returned in about 2.4 ms
> 
> 
> @luukvanderduim
> 
> > What I would like to know is which application is not responding to the `GetApplicationBusAddress` method call during initialization.
> > Is the application exclusively non-responsive on `GetApplicationBusAddress` or does it not respond to other calls either?
> 
> In this reproduction, the first registry entry was `:1.0`, identified as `xfce4-session`. It answered the accessibility-interface queries and `GetApplicationBusAddress`, returning an advertised Unix socket. Construction stalled only after that address had been returned, while creating the peer. A direct introspection request to the advertised socket also did not return within five seconds.

I was on XFCE when I wrote P2P so I am surprised.
 
> So my earlier identification of VS Code as the unresponsive application was too strong: VS Code was involved in the original user operation, but this minimal reproduction points to `xfce4-session`. `GetApplicationBusAddress` itself is not the call hanging here. I have not yet isolated the exact await inside peer creation, so I cannot say whether the stall is during P2P connection/authentication or a subsequent method call.
> 
> > If we can, we should probably just use `Builder::method_timeout` [...] Then we close that failure mode entirely, hopefully and it is a MUCH simpler fix.
> 
> That is an appealing, much smaller option. One caveat is that zbus applies `method_timeout` to method calls after a `Connection` has been built; the socket connection and authentication in `Builder::build()` happen before the timeout is installed on the connection. Therefore it would cover the failure if the stalled operation is a method call, but may not cover a stall during P2P connection/authentication. We need to isolate that exact step before claiming either approach closes the whole failure mode.

I assumed - root of all evil - that it was `GetApplicationBusAddress` that failed to return.

But if the connection builder cannot build, we could use a timeout adapter on the build.
There is something odd going on though because you first identified VS Code as cause but then `xfce-session`, we know and tried both and have not seen the problems you share.
Is there something special in how you run the reproducer? Are they containerized of some sort?

Just for diagnostic purposes, could you try without P2P? Have you tried in a non-Xfce session?

serge-medvedev:
@luukvanderduim

> Is there something special in how you run the reproducer? Are they containerized of some sort?

No. Both binaries were run natively on the same Arch Linux host, as the same user, against the same active Xfce desktop and accessibility bus. They were built from separate detached worktrees. The reproducer itself only calls `AccessibilityConnection::new()`; an external `timeout 10s` distinguishes a return from a stall. No container or VM was involved.

> Just for diagnostic purposes, could you try without P2P?

I have not run upstream with P2P disabled. This branch returning from construction is consistent with not attempting P2P during construction, but that is not the same experiment as building or running upstream without P2P.

> Have you tried in a non-Xfce session?

No, I have only run this comparison in the active Xfce session.

What the comparison established in that one session is limited to:

- upstream `main` did not return within 10 seconds;
- this branch returned in about 2.4 ms;
- upstream reached `xfce4-session` first;
- `GetApplicationBusAddress` returned its advertised Unix socket;
- construction stalled afterward, during peer creation;
- a separate direct introspection attempt against that socket also did not return within five seconds.

I have not isolated the exact await inside peer creation, tested a timeout adapter around `Builder::build()`, reproduced it in another desktop session, or conclusively tied the original observed hang to VS Code. So a timeout around the build is a plausible smaller fix, but I do not yet have test evidence showing whether it covers this failure or what timeout would be appropriate.

PR #383, opened 2026-08-18
fix(common): map the AT-SPI null path to ObjectRef::Null in wire conversions

robin-simular:
## Problem

AT-SPI uses the path `/org/a11y/atspi/null` to mean "no object". Not all of
our conversions handled it:

- `Deserialize` mapped it to `ObjectRef::Null`, but zbus property getters
  don't use `Deserialize` — they use `TryFrom<Value>` / `TryFrom<OwnedValue>`,
  which built a normal ref instead.
- So a null parent came through as a ref that looks real but isn't.
  `is_null()` returns false for it, and calling a method on it sends a
  message with an empty destination. dbus-daemon responds by closing the
  connection.
- Deserializing an empty name with a valid path hit an `assert!`, so bad
  wire data could panic the library.

## Fix

One private constructor now owns the rule "null path means `ObjectRef::Null`".
Every wire conversion goes through it: `Deserialize`, `TryFrom<Value>`,
`TryFrom<OwnedValue>`, and the public constructors. The empty-name case is
now an error instead of a panic.

Two places intentionally skip the rule: header conversions and
`EventProperties::object_ref`. They answer "who sent this event", and a
signal sent from the null path still has a real sender. Both are commented,
and the `_unchecked` constructors also skip the rule (documented).

### DIFF
--- a/atspi-common/src/events/traits.rs
+++ b/atspi-common/src/events/traits.rs
@@ -41,7 +41,9 @@ pub trait EventProperties {
 	fn sender(&self) -> UniqueName<'_>;
 	fn path(&self) -> ObjectPath<'_>;
 	fn object_ref(&self) -> ObjectRef<'_> {
-		ObjectRef::new(self.sender(), self.path())
+		// Reconstruct the event source, not a wire `(so)` pair. A signal
+		// emitted from `/org/a11y/atspi/null` still has a real sender.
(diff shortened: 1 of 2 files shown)


robin-simular:
The CI failure seems unrelated to my changes, but unfortunately it causes some of the tests to be skipped

